PR #5, opened 2025-10-21
Fix build instructions

Yilmaz4:
- The clone link was wrong so I fixed that.
- Without nightly Rust, I kept getting error E0658, so I added instructions to switch to nightly Rust.
- The `cd wayle` command only moves us to the root directory, not the directory that contains the package manifest (`wayle/wayle`), so I fixed the install command to account for that.

(I accidentally opened a pull request for the wrong branch of my fork, oops)

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,11 +117,16 @@ wayle config set general.theme dark
 ```
 
 ## Building
-
+Switch to nightly Rust:
 ```bash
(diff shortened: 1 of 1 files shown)


Jas-SinghFSU:
Ah good catch. I restructured the project when I moved each service into it's own individual crate which changed the build process. Thank you for creating this MR.

PR #19, opened 2026-03-07
fix: also load icons from filepaths for the systray

jasperges:
Some applications provide icon data through a direct file path. This PR adds support for this.

If you want to deal with this differently, please let me know!


Jas-SinghFSU:
Ah nice, good catch. Thank you!

PR #22, opened 2026-03-08
Created NixOS Package & Module

isaacST08:
# This will not be merged

Nix support will be added directly to nixpkgs, without flake.nix in this repo. See https://github.com/NixOS/nixpkgs/pull/503416

---

Created a NixOS package and home-manager module so that wayle can be easily installed and configured using NixOS.

## What was added:

- **NixOS Package:**
  - Provides both the `wayle` and `wayle-shell` cli commands. 
  - Installs all the icons required by wayle and puts them in the correct locations for nix to use them.
  - Is set up so that it can be built independently of the repo which will facilitate easy transfer of the package to the main nixpkgs repo when the time comes.
- **Nix Home-Manager Module:**
  - Allows configuring wayle via nix config in one's `home.nix` file.
  - Has an option to setup wayle as a systemd service that starts wayle automatically.
- **Nix Flake:**
  - A Nix flake was created that supplies the most up to date version of wayle and supplies the nix wayle package to be built with rust nightly (this could change in future but for now it follows the regular build instructions).
  - Also supplies the home-manager module in an easy to follow method.
- **Readme:**
  - The `README.md` file was updated to include how to install wayle by using the nix flake. 
  - Since this new NixOS section is quite large, and not relevant to the large majority of users (I think?), it was formatted into a drop-down as to not bloat the rest of the README. This can be changed if you prefer.

## Going forward:

I am almost certain that this will package will break as wayle is developed simply due to how early into development wayle is. Regardless, I thought that it would still be a good idea to get this started sooner rather than later.

If you do accept this PR, please do not hesitate from doing something that will improve wayle or adding a dependency
with the fear of breaking this nix package/flake. This packaging can be updated and, should something break, you are more than welcome to contact me requesting that I fix it.

### DIFF
--- a/nix/modules/home-manager.nix
+++ b/nix/modules/home-manager.nix
@@ -132,7 +132,7 @@ in
           ConditionEnvironment = "WAYLAND_DISPLAY";
         };
         Service = {
-          ExecStart = "${getExe' cfg.package "wayle-shell"}";
+          ExecStart = "${getExe' cfg.package "wayle"} shell";
           Restart = "on-failure";
         };
         Install = {


PerchunPak:
Okay, I was wrong about this being unusable. I tested this a couple month ago when wayle did not have pop-ups implemented. But I am still not sure if this project is ready. @Jas-SinghFSU, is it ready for an average user?

PerchunPak:
I will start daily-driving this anyway

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,81 @@
...
+} @ s: let
+  inherit (lib.modules) mkIf;
+  inherit (lib.options) mkEnableOption mkOption mkPackageOption;
+
+  cfg = config.services.wayle;
+  tomlFormat = pkgs.formats.toml {};
+in {
+  options.services.wayle = {
+    enable = mkEnableOption "wayle shell";
+    package = mkPackageOption pkgs "wayle" {nullable = true;};
```
PerchunPak: There is not much sense in making package nullable, at least I didn't see anything like that in home-manager codebase

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,81 @@
...
+      description = ''
+        Standard configuration options for wayle.
+      '';
+
+      default = {};
+
+      example = {
+        styling = {
+          theme-provider = "wayle";
+
```
PerchunPak, line 52: This will probably raise an error when trying to render docs; [home-manager uses `lib.literalExpression`](https://github.com/nix-community/home-manager/blob/bb014746edb2a98d975abde4dd40fa240de4cf86/modules/programs/hyprpanel/default.nix#L29)

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,81 @@
...
+    systemd.enable =
+      (mkEnableOption "wayle systemd service")
+      // {
+        default = cfg.enable;
+        example = false;
+      };
+
+    settings = mkOption {
+      type = lib.types.attrs;
```
PerchunPak: It is possible to define the option type as TOML

https://github.com/nix-community/home-manager/blob/bb014746edb2a98d975abde4dd40fa240de4cf86/modules/programs/zk.nix#L23

nix/package.nix
```diff
@@ -0,0 +1,109 @@
...
+    rev = "b4ade2c55c59f4a706192ff784ad11ba38517158"; # Once this ^, then delete this <.
+    hash = "sha256-oc1EtpzH9KLKm0NUpqqQrfbYXftqjVvEfw72QbxlIfc=";
+    fetchSubmodules = true;
+  };
+
+  cargoLock.lockFile = "${finalAttrs.src}/Cargo.lock";
+  cargoHash = "sha256-Y/R8iI1sIHZxqxlPO1iTIqpLJGkCuTPAmh2OarvkoVA=";
+
+  LIBCLANG_PATH = lib.makeLibraryPath [llvmPackages.libclang];
+  LD_LIBRARY_PATH = lib.makeLibraryPath [
```
PerchunPak, line 38: It is discouraged to set `LD_LIBRARY_PATH` directly in Nix, you should instead put dependencies into [`buildInputs`](https://nixos.org/manual/nixpkgs/unstable/#var-stdenv-buildInputs) or [`nativeBuildInputs`](https://nixos.org/manual/nixpkgs/unstable/#var-stdenv-nativeBuildInputs) (the former is for runtime dependencies and the second one is for build-only time dependencies)

nix/package.nix
```diff
@@ -0,0 +1,109 @@
...
+    pkg-config
+  ];
+
+  checkFlags = [
+    # This test is broken because gtk fails to initialize in the nix sandbox.
+    "--skip=css_loads_into_gtk4"
+
+    # This test is broken because wayle is not given access to hyprland in the
+    # nix sandbox.
+    "--skip=new_fails_when_hyprland_instance_signature_missing"
```
PerchunPak, line 69: I am not sure why, but that test passes in my package

nix/package.nix
```diff
@@ -0,0 +1,109 @@
...
+  preBuild = ''
+    # Cargo will fail without a home dir so give it a temporary one and we will
+    # take out what we need from it later.
+    export HOME=$TMPDIR
+
+    # Add clang, glib, and other headers to bindgen search path.
+    export BINDGEN_EXTRA_CLANG_ARGS="$(< ${stdenv.cc}/nix-support/libc-crt1-cflags) \
+      $(< ${stdenv.cc}/nix-support/libc-cflags) \
+      $(< ${stdenv.cc}/nix-support/cc-cflags) \
+      $(< ${stdenv.cc}/nix-support/libcxx-cxxflags) \
```
PerchunPak, line 81: There is [`rustPlatform.bindgenHook`](https://github.com/NixOS/nixpkgs/blob/b314521edc6a41029806005d97ed11f57402ee50/pkgs/build-support/rust/hooks/rust-bindgen-hook.sh). If you are not sure how to do something, try grepping nixpkgs. `bindgen` is definitely used in at least one other package, so you should find the solution there. That applies to pretty much everything

```
~/dev/nixpkgs $ rg --files-with-matches buildRustPackage | xargs rg bindgen
<301 results>
```

nix/package.nix
```diff
@@ -0,0 +1,109 @@
...
+      ${lib.optionalString stdenv.cc.isClang "-idirafter ${stdenv.cc.cc}/lib/clang/${lib.getVersion stdenv.cc.cc}/include"} \
+      ${lib.optionalString stdenv.cc.isGNU "-isystem ${stdenv.cc.cc}/include/c++/${lib.getVersion stdenv.cc.cc} -isystem ${stdenv.cc.cc}/include/c++/${lib.getVersion stdenv.cc.cc}/${stdenv.hostPlatform.config} -idirafter ${stdenv.cc.cc}/lib/gcc/${stdenv.hostPlatform.config}/${lib.getVersion stdenv.cc.cc}/include"}"
+  '';
+
+  postInstall = ''
+    # Generate wayle icons and copy them to the correct nix directory.
+    mkdir -p $out/share
+    $out/bin/wayle icons setup
```
PerchunPak, line 89: Jas said that this command just does `cp -r $src/resources/icons ~/.local/share/wayle/icons`; not to blame you for not knowing though

https://discord.com/channels/1277526090516135978/1277527251189240000/1470198152312262906

README.md
```diff
@@ -184,6 +184,172 @@ wayle icons setup
...
+   cp ~/.config/wayle/config.toml{.bak,}
+   ```
+
+2. Modify the config file at `~/.config/wayle/config.toml` either manually or by
+   using the `wayle config <cmd>` commands until you are happy with your
+   configuration.
+
+3. Translate your new config from the TOML syntax of the
+   `~/.config/wayle/config.toml` file into nix syntax in your `home.nix` file at
+   `services.wayle.settings = { # settings go here }`.
```
PerchunPak, line 323: BTW you can automate this using

```nix
$ nix repl
nix-repl> :p builtins.fromTOML (builtins.readFile ./config.toml)
{
  bar = {
    layout = [
      {
        center = [ "window-title" ];
        left = [
          "hyprland-workspaces"
          "cava"
          "media"
        ];
        monitor = "*";
      }
    ];
  };
}
```

README.md
```diff
@@ -184,6 +184,172 @@ wayle icons setup
...
+5. Rebuild your home-manager config and the new settings will be applied.
+
+This method allows you to configure wayle efficiently and benefit from its live
+config reloading and command line features, while still declaratively writing
+your config using nix.
+
+**CAUTION:** Home manager will automatically move any existing config to a
+`config.toml.hm-bak`, or similar, backup file name. This is expected, however,
+wayle will automatically create a _new_ `config.toml` file if you moved the
+symlink make by home-manager and didn't move it back. If this occurs,
```
PerchunPak, line 340: For me, home-manager just crashes and doesn't update symlinks at all. Maybe there is a setting?

README.md
```diff
@@ -184,6 +184,172 @@ wayle icons setup
...
+```
+
+#### Notes on using NixOS
+
+##### Icons
+
+This nix package will automatically install the icons for you so you don't need
+to worry about running `wayle icons setup`.
+
+##### Live Config Reloading
```
PerchunPak, line 302: I would consider this section a bit redundant since this is unfortunately the case for every software in Nix, and it cannot be fixed. Thus, every user already knows how to work around that. If the user already managed to climb Nix's learning curve...

Image

nix/package.nix
isaacST08: Yeah I didn’t know this, I just thought I’d follow the build instructions in case something special was going on.

README.md
isaacST08: This is amazing to know, thank you.

README.md
isaacST08: I will look into this.

README.md
isaacST08: Yes I suppose that is true. I think I was possible just trying to soften that curve a little bit, even if it was just for a handful of people.

That graphic is also fantastic.

nix/package.nix
isaacST08: Ok, good to know that it is discouraged. I remember something happening that meant that even if the package was in either of the build inputs, it still failed until I added it to `LD_LIBRARY_PATH`. I will look into your package implementation and see if I can find what you did differently to avoid this.

nix/package.nix
isaacST08: Again, I will look into your package to see what you’ve done differently. Maybe it is caused by something else existing or not existing on my machine compared to yours.

nix/package.nix
isaacST08: Just as a side note, I started with a nix shell and then went to a package. I started by following https://nixos.wiki/wiki/Rust#Installation_via_rustup . They use `LD_LIBRARY_PATH`, is this only acceptable because it is a shell or should the wiki get a note mentioning this?

PerchunPak: Yes, it is allowed in nix shells. `buildInputs` make sense only for packages, since you don't build anything in a shell, just putting packages into $PATH

nix/package.nix
PerchunPak: > Maybe it is caused by something else existing or not existing on my machine compared to yours.

Nix is notoriously strong about enforcing reproducibility, so it is pretty impossible. Unless drivers, kernel or something similar mess up something (which is not the case here). 

Packages in `nativeBuildInputs` may register some hooks, so it is probably the reason (you put most packages in the `LD_LIBRARY_PATH` while I used `nativeBuildInputs`).

README.md
isaacST08: I have home-manager configured as a module in my main flake and I have the following option set in its config:
```nix
# Inside "inputs.nixpkgs.lib.nixosSystem {...}":
modules = [
  ...
  {
    inputs.home-manager.nixosModules.home-manager 
    {
      home-manager = {
        ...
        backupFileExtension = "hm-bak";
        ...
      };
    }
  }
  ...
```
This must be the difference that causes your homemanager to fail to take over and generate the symlinks. However I don't have an easy solution for users who don't want to set this option because wayle seems to automatically generate a new file at any opportunitiy which will need either backing up or removing manually without the option.

nix/package.nix
isaacST08: As mentioned below, this test is still failing on my system even when I use your package.

isaacST08:
> Yeah, the test is broken on main right now. I am using an older version of wayle currently. I have pushed a new version, but didn't test the build (will do it later today)
> 
> [PerchunPak/nixpkgs@2a9b6ac](https://github.com/PerchunPak/nixpkgs/commit/2a9b6ac1dedbb3cd72ee6f4157910930e93fd301)
> 
> EDIT: I am talking about `types::tests::screencast_owner_try_from_fails_for_invalid_value`, the other one doesn't fail for me

I've just tried this new commit and it successfully built. I'm getting some other errors after running wayle but I don't yet know if those are nix related yet.

PerchunPak:
> I've just tried this new commit and it successfully built. I'm getting some other errors after running wayle but I don't yet know if those are nix related yet.

Are you sure? That commit didn't skip 3 other failing tests. Jas fixed those today and this commit definetly compiles (at least on ARM, compiling for x86 right now)

https://github.com/PerchunPak/nixpkgs/commit/e74327a46103e9b01f2951b8cdb4924ecb704465

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,147 @@
...
+  cfg = config.services.wayle;
+
+  tomlFormat = pkgs.formats.toml { };
+in
+{
+  meta.maintainers = with lib.maintainers; [ isaacST08 ];
+
+  options.services.wayle = {
+    enable = mkEnableOption "wayle shell";
+    package = mkPackageOption pkgs "wayle" { nullable = true; };
```
PerchunPak: There are nullable packages in home-manager, but I am not convinced it is useful here. Only adds complexety to the code

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,147 @@
...
+
+    systemd.enable = (mkEnableOption "wayle systemd service") // {
```
PerchunPak: The same here, it is in the hyprpanel module only for backwards compatibility. User can always disable this using `systemd.user.services.wayle.Disable` (or something like that, don't remember the option name)

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,147 @@
...
+          "wallust"
+          "pywal"
+        ]) pkgs.${themeProvider})
+      )
+    );
+
+    # Main config file.
+    xdg.configFile."wayle/config.toml" = mkIf (cfg.settings != { }) {
+      source = tomlFormat.generate "wayle-config" cfg.settings;
+      force = true; # Wayle aggressively produces its own config file.
```
PerchunPak: It used to be true for hyprpanel, but I did not encounter this problem with wayle
```suggestion
```

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,147 @@
...
+
+      # Systemd service to run the wallpaper cycling.
+      wayle-wallpaper =
+        mkIf
+          (
+            attrByPath [ "wallpaper" "cycling-enabled" ] false cfg.settings
+            && "" != attrByPath [ "wallpaper" "cycling-directory" ] "" cfg.settings
+          )
+          {
+            Unit.Description = "Wayle - Wallpaper Cycling";
```
PerchunPak: This seems to work fine without a systemd unit, I would expect it to conflict with wayle's implementation

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,147 @@
...
+  inherit (lib.attrsets) attrByPath hasAttrByPath;
+  inherit (lib.modules) mkIf;
+  inherit (lib.options) mkEnableOption mkOption mkPackageOption;
+
+  cfg = config.services.wayle;
+
+  tomlFormat = pkgs.formats.toml { };
+in
+{
+  meta.maintainers = with lib.maintainers; [ isaacST08 ];
```
PerchunPak:
```suggestion
  meta.maintainers = with lib.maintainers; [
    isaacST08
    PerchunPak
  ];
```

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,147 @@
...
+      (lib.lists.optional (cfg.package != null) cfg.package)
+      # Install the appropriate theme-provider, if set.
+      ++ (
+        let
+          themeProvider = attrByPath [ "styling" "theme-provider" ] "wayle" cfg.settings;
+        in
+        (lists.optional (elem themeProvider [
+          "matugen"
+          "wallust"
+          "pywal"
```
PerchunPak: wallust and pywal have home-manager modules, we should probably enable them instead of just adding them to `home.packages`

nix/modules/home-manager.nix
isaacST08: Yeah I totally agree. I wanted to remove it but I just thought that maybe just maybe there might be someone who didn't want to actually install the wayle package via nix but did want to have its configuration generated by nix. If you think that is too esoteric of a use case then I would be more than happy to remove it.

PerchunPak: I do. Though there is no established pattern where it is used in home-manager, I don't think someone would do that

nix/modules/home-manager.nix
isaacST08: I've personally had this problem quite a bit. I don't like the solution of forcing either. I think if we were to not use the force flag then we would need an assertion that the file isn't already there, accompanied by a very clear error message for when (I don't think it's a matter of "if") it fails.

nix/modules/home-manager.nix
isaacST08: Interesting... Mine does not work at all no matter how I configure wayle or how many times I try and reset it. Maybe another issue on wayle's side?

nix/modules/home-manager.nix
PerchunPak: > I've personally had this problem quite a bit.

Did you use `wayle config set`?

> I think if we were to not use the force flag then we would need an assertion that the file isn't already there, accompanied by a very clear error message for when (I don't think it's a matter of "if") it fails.

Unfortunately, this is not possible. home-manager tries to get a nice error message, but it is only accessible in the journalctl logs, which is IMO not an obvious place to find an explanation

nix/modules/home-manager.nix
PerchunPak: There is a bug [here](https://github.com/Jas-SinghFSU/wayle/blob/cbccee657aee883697d1b8ae3aa5be91392ca101/crates/wayle-wallpaper/src/backend/awww/mod.rs#L25), wayle just can't find the swww binary. You need to create an alias, just put this into your `home.packages`:

```nix
(writeShellScript "awww" ''
  exec swww "$@"
'')
```

nix/modules/home-manager.nix
isaacST08: Yes, I'm aware. I don't know how they might conflict with each other though and I found that they seemed to work sufficiently with just the package.

It might be required that either the settings from the wallust/pywal modules get translated and applied to the settings for wayle, or vise versa. I also don't know which direction is best.

Additionally, I don't know if when the wallust or pywal home-manager modules are enabled if they even need to be enabled in wayle's config at all. TBH, I don't know exactly how wayle interfaces with these tools so some experimentation might be required.

Additionally, there should probably be a way to let the user choose the package that gets installed for matugen but just default to using the one from `pkgs`.

nix/modules/home-manager.nix
isaacST08: I don't see you in the home-managers maintainers list, so you'll need to make sure you add your information to that. 

I turned on issues on my home-manager fork so feel free to add any comments there directly or make a PR so that your git contributions are actually recorded.

nix/modules/home-manager.nix
isaacST08: > Did you use `wayle config set`?

Nope. Wayle has just been really trigger happy at making a new, blank config file. Granted, it usually happens when I move/copy the symlink so that I can edit the config without rebuilding nix. But I know that the issue is so I can quickly fix it. But I would really like to remove any potential headaches from other nix users if possible.

> Unfortunately, this is not possible. home-manager tries to get a nice error message, but it is only accessible in the journalctl logs, which is IMO not an obvious place to find an explanation

I was afraid this would be the case.

I don't have a good solution for this then. I think the only thing left would be to let the user decide whether to enable to force flag or not. I guess it's just a question of what to make the default for wayle and use `mkDefault` if we decide the default should be to force.

nix/modules/home-manager.nix
isaacST08: I assume you meant `writeShellScriptBin`, but regardless, yes this worked. Thank you.

I applied the changes and the second systemd service is, as you said, unnecessary.

nix/modules/home-manager.nix
PerchunPak: I am actually

https://github.com/nix-community/home-manager/blob/c6fe2944ad9f2444b2d767c4a5edee7c166e8a95/all-maintainers.nix#L345-L356

PerchunPak: > I turned on issues on my home-manager fork so feel free to add any comments there directly or make a PR so that your git contributions are actually recorded.

The only contributions that count towards getting permission to merge PRs are created and reviewed PRs, so it doesn't really matter.

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,144 @@
...
+          Restart = "on-failure";
+        };
+      };
+
+      # Wallpaper-engine dependency.
+      services.swww.enable = mkIf settings.wallpaper.engine-enabled (lib.mkDefault true);
+
+      # If wallust or pywal is the theme-provider and is enabled globally,
+      # ensure the theme gets sourced for new terminals.
+      programs.bash.bashrcExtra =
```
PerchunPak: There are shells other than bash, and it is something that should be done in the `pywal`/`wallust` modules.

nix/modules/home-manager.nix
PerchunPak: > Nope. Wayle has just been really trigger happy at making a new, blank config file. Granted, it usually happens when I move/copy the symlink so that I can edit the config without rebuilding nix. But I know that the issue is so I can quickly fix it. But I would really like to remove any potential headaches from other nix users if possible.

I have two commands just for that

```nix
(pkgs.writeShellScriptBin "modify" ''
  set -ex
  mv "$1" "$1"1
  cat "$1"1 > "$1"
'')

(pkgs.writeShellScriptBin "modifyu" ''
  set -ex
  mv "$1" "$1".modified
  mv "$1"1 "$1"
'')
```

```sh
$ modify config.toml
+ mv config.toml config.toml1
+ cat config.toml1 > config.toml
$ modifyu config.tom
+ mv config.toml config.toml.modified
+ mv config.toml1 config.toml
```

It is really annoying how long nix rebuilds are. Anyway, we don't want users to lose their modified configs accidentally, that's why `force` is off by default

nix/modules/home-manager.nix
PerchunPak: You are right, `pywal` module does some additional stuff (adds integration to kitty, xterminal, etc). Adding binaries is enough. Though it would be good to have an `autoInstallDependencies`. It is hard to remove something from a list in nixos modules

Actually, the proper way to do this is create a wrapper around wayle with additional dependencies, but it is very complicated and inconvenient. I don't know a good way to do that yet

nix/modules/home-manager.nix
isaacST08: Agreed. This should be a separate PR on those modules. I’ll remove it.

nix/modules/home-manager.nix
isaacST08: > I am actually

My appologies, I didn't realize that there were two lists, I only searched in `modules/lib/maintainers.nix`. My bad.

nix/modules/home-manager.nix
```diff
@@ -0,0 +1,137 @@
...
+
+      # Systemd service for main wayle shell.
+      systemd.user.services.wayle = {
+        Unit.Description = "Wayle - Shell";
+        Install.WantedBy = [ "graphical-session.target" ];
+        Service = {
+          ExecStart = "${getExe' cfg.package "wayle-shell"}";
+          Restart = "on-failure";
+        };
+      };
```
PerchunPak, line 141: For some reason, wayle does not automatically start for me. Maybe steal the systemd unit from hyprpanel?

```suggestion
      systemd.user.services.wayle = {
        Unit = {
          Description = "Wayland Elements - A compositor agnostic shell with extensive customization";
          Documentation = "https://github.com/Jas-SinghFSU/wayle";
          PartOf = [ config.wayland.systemd.target ];
          After = [ config.wayland.systemd.target ];
          ConditionEnvironment = "WAYLAND_DISPLAY";
        };

        Service = {
          ExecStart = "${getExe' cfg.package "wayle-shell"}";
          Restart = "on-failure";
        };

        Install = {
          WantedBy = [ config.wayland.systemd.target ];
        };
      };
```

isaacST08: Should the documentation path instead be changed to `https://github.com/wayle-rs/wayle`? Or do you know another something that I don't know?

PerchunPak: Yes, it should be

PerchunPak:
> @PerchunPak if everything looks good to you, I’ll open a draft PR for home-manager later today.

Everything looks good and works; thank you for your work. Yes, please open a PR

PerchunPak:
> @PerchunPack if I could request a little bit of advice since you've contributed to home-manager before: One of the checks is running `nix run .#tests -- test-all`, however, this fails. The tests I wrote for wayle pass (running `nix run .#tests -- wayle), but when running the full set, something related to podman fails. This same set of errors also occurs when running `nix run .#tests -- test-all` on the master branch.
> 
> Should I still create the PR as a draft regardless? Any hints as to why the tests even on master would be failing?

Yes, create the PR. Running all tests is unnecessary, and those are very heavy; I don't know why running everything instead of things that were touched is in the PR template.

Jas-SinghFSU:
> > @PerchunPack if I could request a little bit of advice since you've contributed to home-manager before: One of the checks is running `nix run .#tests -- test-all`, however, this fails. The tests I wrote for wayle pass (running `nix run .#tests -- wayle), but when running the full set, something related to podman fails. This same set of errors also occurs when running `nix run .#tests -- test-all` on the master branch.
> > Should I still create the PR as a draft regardless? Any hints as to why the tests even on master would be failing?
> 
> Yes, create the PR. Running all tests is unnecessary, and those are very heavy; I don't know why running everything instead of things that were touched is in the PR template.

I've been meaning to get to that for GH actions. My baby-steps to getting there so far have failed.

PR #28, opened 2026-03-10
feat(keyboard-input): add support for overriding language names

### LINKED ISSUE #23: Country flags for languages
This is a nice feature that KDE has, basically, instead of displaying `English (US)`, the interface displays 🇺🇸. Mapping every possible language to a flag is very time-consuming, but it would be nice to expose something like a `language-override` option

```toml
[modules.keyboard-input]
language-override = {
  # replaces "English (US)" with 🇺🇸
  "English (US)" = "🇺🇸",
  "CZ (Qwerty)" = "🇨🇿",
  "UA" = "🇺🇦",
}
```

PerchunPak:
Closes https://github.com/Jas-SinghFSU/wayle/issues/23

This adds a new config option, which allows replacing language names with custom text. This is useful for keyboard layouts that have additional stuff in them (like `Czech (QWERTY)` or `English (US)`) or replacing languages with abbreviations, so switching layouts doesn't trigger jumps.

```toml
[modules.keyboard-input.language-name-map]
"English (US)" = "EN"
"Czech (QWERTY)" = "Czech"
```

Also tested with my usecase (https://github.com/Jas-SinghFSU/wayle/issues/23), but the text now is very small (would be nice to be able to configure paddings/text-font for individual modules)

image

---

Had to refactor `HyprlandKeyboardInput.update_label` a bit so it doesn't get layout as an argument, but instead always gets it from the config. Providing it each time would be annoying. I hope this doesn't break anything


crates/wayle-config/src/schemas/modules/keyboard_input/mod.rs
```diff
@@ -14,12 +16,14 @@ pub struct KeyboardInputConfig {
...
+    /// - `{{ layout }}` - Raw layout name from the compositor (e.g., "English (US)")
+    /// - `{{ alias }}` - User-defined alias from `layout-alias-map`, falls back to`{{ layout }}`
     ///
     /// ## Examples
     ///
-    /// - `"{{ layout }}"` - "us"
-    #[default(String::from("{{ layout }}"))]
+    /// - `"{{ layout }}"` - "English (US)"
+    /// - `"{{ alias }}"` - "EN" (with alias map configured)
+    #[default(String::from("{{ alias }}"))]
```
Jas-SinghFSU, line 26:
```suggestion
    #[default(String::from("{{ layout }}"))]
```
I think we should stick with layout as the default and make aliases opt in.

crates/wayle-config/src/schemas/modules/keyboard_input/mod.rs
```diff
@@ -14,12 +16,14 @@ pub struct KeyboardInputConfig {
     ///
     /// ## Placeholders
     ///
-    /// - `{{ layout }}` - Keyboard layout code (e.g., "us", "de", "fr")
+    /// - `{{ layout }}` - Raw layout name from the compositor (e.g., "English (US)")
+    /// - `{{ alias }}` - User-defined alias from `layout-alias-map`, falls back to`{{ layout }}`
```
Jas-SinghFSU:
```suggestion
    /// - `{{ alias }}` - User-defined alias from `layout-alias-map`, falls back to `{{ layout }}`
```

Jas-SinghFSU: Nit: Missing space between `{{ layout }}` and 'falls back to'

Jas-SinghFSU (changes requested):
Just a couple of small changes. Looks great! Thanks for making the changes.

crates/wayle-config/src/schemas/modules/keyboard_input/mod.rs
PerchunPak: Oh, sorry, forgot to leave a comment about that.

Isn't it confusing to have an option that doesn't do anything by default? `{{ alias }}` falls back to `{{ layout }}` anyway, so there is no harm in having it by default.

I only see one use case for a separate `{{ alias }}` variable: `format = "{{ layout }} | {{ alias }}"` (using them both at the same time instead of one of them). There is no need to use `format = "{{ layout }}"` because the user fully controls what is in `{{ alias }}`

Jas-SinghFSU: Fair points, I'm aligned with that.

PR #32, opened 2026-03-13
Fix icon location in the resources directory

jasperges:
When running `wayle icons setup` it copies the icons from `resources/icons/hicolor/scalable/actions`. Some icons were directly in `resources/icons` and were not copied over. This resulted in some broken icons, for example in the dashboard. Files that were already the `actions` directory are simply removed from `resources/icons`.


Jas-SinghFSU:
Thank you!

PR #44, opened 2026-03-22
feat(shell): Sleep and hibernate buttons

justjustie:
I have also added buttons for sleep and hibernation:
grafik

~~Layout is done using a grid to achieve the two rows, which feels kind of hacky but I've never done gtk before :P gladly let me know if that needs improvement.~~ Nevermind got it

I am also looking for feedback on order and color coding of the buttons. I especially wonder if you consider sleep/hibernation unrecoverable as they also need user interaction to come back from (power button on my hardware).

Let me know what you think!!


justjustie:
The buttons got a little small in the process, so I've added some padding to them:
grafik

isaacST08:
> What do we think about this kinda layout?

I think for the number of buttons it's pretty good. I do think however that if the user configures to use ~4 or less buttons then it should default back to the original side-by-side layout. However this might have to change based on how long the user's username is.
So I guess, again, maybe just yet another config option.

But for a top-over-bottom layout, I like it. Maybe just a little bit more horizontal spacing between the buttons though.

justjustie:
Agreed on the spacing.

As for the other things, I wonder if that could be an entire issue on its own maybe. For example I am on desktop and therefore the WiFi, BT and airplane mode buttons are just permanently disabled, I'd love to be able to hide them too.

Jas-SinghFSU:
I think this looks nice, though one major concern that I have is that the dashboard dropdown is already quite large vertically. This introduces additional verticality, thus making the dropdown even larger.

Perhaps we keep it at 4 buttons, but allow those buttons to be fully configurable in terms of:
1. Enabling/disabling it
2. Changing the icon + tooltip
3. Changing the command it executes

So that you have less button bloat while still providing the flexibility.

justjustie:
Sounds good to me too. I don't think what I've done in this PR helps that direction really, and neither would I currently be able to implement this, so I'd suggest closing this and writing a new issue, what do you think?

While were on this topic, a setting to hide the topmost 3 buttons on the dashboard (WiFi, BT, Airplane) would also be useful IMO, as they are not needed on your average desktop PC.

Jas-SinghFSU:
Yeah, dropdown configurability is probably warranted. We can close this then add handle it as a different PR. Thanks for your efforts!

PR #51, opened 2026-03-24
configurable battery and volume labels

Dieterbe:
i'm still working on this. running into some errors still.
just publishing to let you know what i'm up to

### DIFF
--- a/crates/wayle-shell/src/shell/bar/factory.rs
+++ b/crates/wayle-shell/src/shell/bar/factory.rs
@@ -124,7 +124,7 @@ fn sync_container_visibility(container: &gtk::Box) {
         .observe_children()
         .into_iter()
         .filter_map(|obj| obj.ok()?.downcast::<gtk::Widget>().ok())
-        .any(|widget| widget.is_visible());
+        .any(|widget| widget.get_visible());
 
     container.set_visible(has_visible_child);
 }


crates/wayle-config/src/schemas/modules/battery/mod.rs
```diff
@@ -72,6 +72,19 @@ pub struct BatteryConfig {
     #[default(ColorValue::Token(CssToken::Yellow))]
     pub label_color: ConfigProperty<ColorValue>,
 
+    /// Format string for the label.
+    ///
+    /// ## Placeholders
+    ///
+    /// - `{{ percent }}` - Baterry level (0-100)
```
Jas-SinghFSU:
```suggestion
    /// - `{{ percent }}` - Battery level (0-100)
```
Typo

crates/wayle-shell/src/shell/bar/modules/volume/watchers.rs
```diff
@@ -22,7 +22,8 @@ pub(super) fn spawn_watchers(
 
     let level_icons = config.level_icons.clone();
     let muted_icon = config.icon_muted.clone();
-    watch!(sender, [level_icons.watch(), muted_icon.watch()], |out| {
+    let format = config.format.clone();
+    watch!(sender, [level_icons.watch(), muted_icon.watch(), format.watch()], |out| {
         let _ = out.send(VolumeCmd::IconConfigChanged);
```
Jas-SinghFSU:
```suggestion
        let _ = out.send(VolumeCmd::ConfigChanged);
```
We should probably take this opportunity to just make this a generic name 'ConfigChanged' since my original is a bit misleading. It updates both the label AND icons.

PR #57, opened 2026-03-25
Add cpu-chart module

Dieterbe:
- factor out CAVA barchart related code to more generic code
- reuse that generic code to power a new cpu-chart module, which shows cpu usage per core as a barchart

config is based on cava, just taking what's relevant for the barchart

here's what it looks like with 1 or 2 of my cores doing some work, on my system, with the below config: (note that i have it in a layout group together with a cpu module which shows avg freq and temp)

```
[modules.cpu-chart]
bar-width = 4
bar-gap = 0
direction = "normal"
internal-padding = 0
button-bg-color = "bg-surface-elevated"
left-click = "xdg-terminal-exec top"
right-click = "xdg-terminal-exec btop"
```

image

I don't have much experience with Rust, but it seems to work fine.
the one thing i didn't pay attention to, is what i need to do to include it when generating a default config, should i do something?


Jas-SinghFSU:
I think a really neat feature would be to allow a configuration setting that lets users pick which cores they want to monitor specifically.

crates/wayle-config/src/schemas/bar/types/mod.rs
```diff
@@ -354,6 +358,7 @@ impl fmt::Display for BarModule {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Custom(id) => write!(f, "{}{}", Self::CUSTOM_PREFIX, id),
+            Self::CpuChart => f.write_str("cpuchart"),
```
Jas-SinghFSU: We can remove this. It's redundant since we've already added it to our to_kebab_case helper. And it'll fall-through to that automatically via:
```rust
            _ => f.write_str(self.to_kebab_case()),
```

crates/wayle-config/src/schemas/bar/types/mod.rs
```diff
@@ -250,6 +252,7 @@ impl BarModule {
             Self::Cava => "cava",
             Self::Clock => "clock",
             Self::Cpu => "cpu",
+            Self::CpuChart => "cpuchart",
```
Jas-SinghFSU:
```suggestion
            Self::CpuChart => "cpu-chart",
```
Let's make it kebab case.

crates/wayle-config/src/schemas/modules/cava/mod.rs
```diff
@@ -63,8 +65,8 @@ pub struct CavaConfig {
     pub style: ConfigProperty<CavaStyle>,
 
     /// Bar growth direction.
-    #[default(CavaDirection::default())]
-    pub direction: ConfigProperty<CavaDirection>,
+    #[default(BarDirection::Normal)]
```
Jas-SinghFSU: Since you've defined a default on the enum:
```rust
    #[default]
    Normal,
```
we can just use
```rust
    #[default(BarDirection::default())]
```

crates/wayle-config/src/schemas/modules/cpuchart.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::{
+    ClickAction, ConfigProperty,
+    schemas::{
+        barchart::BarDirection,
+        styling::{ColorValue, CssToken, Spacing},
+    },
+};
+
+/// Configuration for the CPU chart module.
+#[wayle_config(bar_container)]
```
Jas-SinghFSU: I see that we're using the `bar_container` template here but in the actual module we're using `bar_button`. Perhaps we use `bar_button` here as well? So that the config schema enforces that we provide the right properties.

crates/wayle-config/src/schemas/modules/mod.rs
```diff
@@ -80,6 +83,9 @@ pub struct ModulesConfig {
     pub clock: ClockConfig,
     /// CPU usage module.
     pub cpu: CpuConfig,
+    /// CPU usage chart module.
+    #[serde(rename = "cpu-chart")]
+    pub cpuchart: CpuChartConfig,
```
Jas-SinghFSU: Nit: Could we use snake_case for the CPU chart module everywhere if we're translating it to kebab-case?

crates/wayle-config/src/schemas/modules/cpu_chart/mod.rs
Jas-SinghFSU, line 1: Convention: Could we move this to `cpu_chart/mod.rs` to follow our existing patterns?

crates/wayle-config/src/schemas/types/barchart.rs
Jas-SinghFSU, line 1: Nit: To keep the schemas/ directory clean, could we move shared types into a new types/ directory. So this would move to `crates/wayle-config/src/schemas/types/barchart.rs`.

crates/wayle-shell/src/shell/bar/modules/cava/rendering/mod.rs
```diff
@@ -1,31 +1,25 @@
...
 
-const MIN_BAR_HEIGHT: f64 = 2.0;
-
-pub(super) struct RenderParams {
-    pub bar_width: f64,
-    pub bar_spacing: f64,
-    pub fill_color: Rgba,
-}
+// Use BarchartParams directly instead of wrapper type
+pub(super) type RenderParams = BarchartParams;
```
Jas-SinghFSU: Hmm... something about this is throwing me off a bit. These are BarChartParams (implying they're for bars) and they exist in the bar chart components, but the same struct is being used in the wave module as well. Perhaps we want to move them elsewhere if they're not specific to bars and change their name?

Dieterbe:
> I think a really neat feature would be to allow a configuration setting that lets users pick which cores they want to monitor specifically.

Why? At least on Linux (but AFAIK on other OS'es also) processes get randomly distributed across all cores and pinning workloads to specific cores is very rarely done, so why treat any particular cores as special?

Jas-SinghFSU:
> > I think a really neat feature would be to allow a configuration setting that lets users pick which cores they want to monitor specifically.
> 
> Why? At least on Linux (but AFAIK on other OS'es also) processes get randomly distributed across all cores and pinning workloads to specific cores is very rarely done, so why treat any particular cores as special?

Yeah, I'm inclined to probably push back on my own suggestion and agree with you. There are scenarios in which you may want to pin processes to a specific core (though you'd probably be better off **not** doing that in 99% of cases - unless you really know what you're doing). In that case you may want to watch cores that you specifically care about.

Though this is more theoretical than practical and I would argue that it's not worth the added complexity for supporting a niche use-case.

crates/wayle-shell/src/shell/bar/modules/cpuchart/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+                        self.is_vertical,
+                    );
+                }
+                self.drawing_area.queue_draw();
+            }
+        }
+    }
+}
+
+fn setup_draw_func(
```
Jas-SinghFSU: Convention: This will be in the contribution guidelines in the future but a convention I use for Relm4 modules is to keep the `mod.rs` file strictly for the relm4 Component impl:
```rust
impl Component for CpuChartModule {
```

Any other logic is split up into the following files:
1. watchers.rs - any long running tasks that watch for changes and emit signals (you've got this one correct)
2. factory.rs - factory register (you've already done this well)
3. messages.rs - the message signature of the component
4. helpers.rs - for any pure and easily unit testable functions
5. methods.rs - any methods or free functions that rely on the framework

So I would recommend that we move all the free functions like setup_draw_func into `methods.rs`. That way `mod.rs` just gives a 1000ft view of the root component.

I can understand that this may be a bit over the top but it helps my fish-brain easily understand modules at a high level by just looking at the mod.rs without the rest of the implementation detail noise.

crates/wayle-shell/src/shell/bar/modules/cpuchart/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+        watchers::spawn_watchers(&sender, cpuchart_config, &init.sysinfo);
+
+        let bar_button = BarButton::builder()
+            .launch(BarButtonInit {
+                icon: String::new(),
+                label: String::new(),
+                tooltip: None,
+                colors: BarButtonColors {
+                    icon_color: ConfigProperty::new(
+                        wayle_config::schemas::styling::ColorValue::Token(CssToken::FgDefault),
```
Jas-SinghFSU: Can we import these at the top so we don't have fully qualified paths in-line and can just do
```rust
ColorValue::Token(CssToken::FgDefault),
```
Additionally, as per my comment in our config section, if we're going to render a BarButton instead of BarContainer, we may as well allow icon section to be configurable.

crates/wayle-shell/src/shell/bar/modules/cpuchart/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        let full_config = config_clone.config();
+        let cpuchart_config = &full_config.modules.cpuchart;
+
+        let bar_width = cpuchart_config.bar_width.get() as f64;
+        let bar_spacing = cpuchart_config.bar_gap.get() as f64;
+        let bar_scale = full_config.bar.scale.get().value();
+        let padding_rem = cpuchart_config.internal_padding.get().value();
+        let horizontal_padding = super::shared::rem_to_px(padding_rem, bar_scale);
```
Jas-SinghFSU: Nit: Can we bring `super::shared` in at the top? Any any other inline import that used more than once should probably be brought in towards the top next to the other super/crate uses.

crates/wayle-shell/src/shell/bar/modules/cpu_chart/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+            is_vertical,
+            dropdowns: init.dropdowns,
+        };
+        let bar_button_widget = model.bar_button.widget();
+
+        if let Some(button_box) = bar_button_widget.child().and_downcast::<gtk4::Box>() {
+            let mut child = button_box.first_child();
+            while let Some(current) = child {
+                if current.css_classes().iter().any(|c| c == "label-container") {
+                    if let Some(label_container) = current.downcast_ref::<gtk4::Box>() {
```
Jas-SinghFSU: Nit: As per clippy, we can reduce this nesting by joining the top level if and this if with `&&`

crates/wayle-shell/src/shell/bar/modules/cpuchart/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+    let total_width =
+        (num_cores as f64 * (bar_width + bar_spacing)) - bar_spacing + (2.0 * horizontal_padding);
+
+    if is_vertical {
+        drawing_area.set_height_request(total_width as i32);
+        drawing_area.set_width_request(20);
+    } else {
+        drawing_area.set_width_request(total_width as i32);
+        drawing_area.set_height_request(20);
+    }
```
Jas-SinghFSU:
```suggestion
    if is_vertical {
        drawing_area.set_size_request(-1, total_width as i32);
    } else {
        drawing_area.set_size_request(total_width as i32, -1);
    }
```
We can just use set_size_request with  -1 so it auto fills it's container.

Jas-SinghFSU (changes requested):
I've noticed that on startup (or when I add this module to the bar), it shows only 1 bar with sizing a little off. I have to remove the module and add it back to fix it.

Dieterbe:
> I've noticed that on startup (or when I add this module to the bar), it shows only 1 bar with sizing a little off. I have to remove the module and add it back to fix it.

I'll look into this. It may be caused by the CPU core measurements themselves. IIRC at startup they are empty first, then only have 1 core and then further updates get all the values. I did not dig into that yet.

crates/wayle-config/src/schemas/modules/mod.rs
Dieterbe: i don't understand. we use kebab-case for all other modules too.
why should it change to snake_case?

crates/wayle-shell/src/shell/bar/modules/cpuchart/mod.rs
Dieterbe: makes sense

crates/wayle-shell/src/shell/bar/modules/cpuchart/mod.rs
Dieterbe: ok, sounds like a good rule of thumb.
inline once is okay, but more than once, bring to top.

crates/wayle-config/src/schemas/modules/mod.rs
Jas-SinghFSU: All rust variables follow the idiomatic snake_case convention. For exposing it to our config.toml (user input), we convert them to kebab-case via:
```rust
    #[serde(rename = "cpu-chart")]
```

I'm noting that we should reflecting the kebab-casiness in our variable name by making it snake case:
```rust
    #[serde(rename = "cpu-chart")]
    pub cpu_chart: CpuChartConfig,
```

Dieterbe: ah yes, got it thanks

PR #63, opened 2026-03-28
Apply color to background shorthand to override theming

### LINKED ISSUE #50: Background is not truly transparent
### Quick check

### Description

Background is not truly transparent even when setting everything possible to the intuitive values.

### Steps to reproduce

1. Set the default config
2. Set: 
```
bar.bg = "transparent"
bar.background-opacity = 0
styling.palette.bg = "#ffffff"
```
3. Start wayle or hot reload.

### Expected behavior

To the bar being transparent, showing the background

### Wayle version

0.1.0

### Distro

Arch Linux

### Compositor and version

Hyprland 0.54.2 built from branch v0.54.2 at commit 59f9f2688ac508a0584d1462151195a6c4992f99 clean (version: bump to 0.54.2).

### Display setup

_No response_

### Relevant config

```toml

```

### Logs

```shell

```

### Affected area

Bar

faculerena:
Fixed #50.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2788,9 +2788,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.12.2"
+version = "1.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "843bc0191f75f3e22651ae5f1e72939ab2f72a4bc30fa80a066bd66edefc24d4"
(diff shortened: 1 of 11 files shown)


Jas-SinghFSU:
I've created #68 so that something like this doesn't happen again in the future for other elements/styles. It'll unset all styling so only Wayle internal styles are retained.

PR #83, opened 2026-03-31
feat: add custom dropdown picker for bar modules

agrahamlincoln:
## Objective

Custom modules can execute shell commands and display output in the bar, but they can't open dropdown panels — the only interaction options are fire-and-forget shell commands or opening built-in dropdowns by name. This means use cases like context switchers (display scale, kubectl context, SSH host picker) require external tools like wofi/rofi for selection UI, which breaks visual consistency with the rest of the panel.

This PR adds config-driven custom dropdown panels with composable sections, starting with a `picker` section type for list-based selection.

Related: #71 explores a similar use case. This takes a different approach — registering custom dropdowns in the `DropdownRegistry` rather than managing popovers directly in the custom module, so they get the same lifecycle behavior as built-in dropdowns (positioning, margins, mutual exclusion, freeze/thaw, keyboard mode).

## Solution

New `[dropdowns.custom.<name>]` config section with a `[[sections]]` array. Currently supports `type = "picker"` — a list-based item picker driven by shell commands:

```toml
[[modules.custom]]
id = "scale"
command = "hypr-scale.sh get"
format = "{{ text }}"
icon-name = "ld-monitor-symbolic"
left-click = "dropdown:custom:scale"
on-action = "hypr-scale.sh get"

[dropdowns.custom.scale]
title = "Display Scale"
icon = "ld-monitor-symbolic"
width = 340

[[dropdowns.custom.scale.sections]]
type = "picker"
list-command = "hypr-scale.sh list --json"
select-command = 'hypr-scale.sh set "$WAYLE_SELECTED"'
```

The `list-command` outputs JSON lines matching the `PopoverItem` fields:

```json
{"value": "1.5", "label": "1.5x", "subtitle": "140 DPI  2560×1707", "active": true}
```

Plain text fallback (one item per line) is also supported for simple cases like `kubectl config get-contexts -o name`.

### Key design decisions

- **Registry integration**: Custom dropdowns register as `"custom:<name>"` in the existing `DropdownRegistry`, getting positioning, margins, mutual exclusion, autohide, freeze/thaw, and keyboard mode for free.
- **Composable sections**: The `[[sections]]` array is designed for future section types (slider, toggle, info) without redesigning the config schema. For now, only `picker` is implemented.
- **`ClickAction` on custom modules**: Changed custom module click fields from `String` to `ClickAction` so `left-click = "dropdown:custom:scale"` works through existing dispatch. Backwards-compatible — `"pavucontrol"` still parses as `Shell(...)`.
- **`$WAYLE_SELECTED` env var**: The selected item's value is passed to `select-command` via environment variable, avoiding shell injection.
- **Post-selection refresh**: After selection, the popover closes and the module's `on-action` runs to update the bar label immediately.
- **Pre-warming**: Custom dropdowns are pre-warmed at startup alongside built-in dropdowns.

### Changes

- `wayle-config`: New `schemas::dropdowns` module with `CustomDropdownDefinition`, `SectionDefinition`, `SectionType::Picker`
- `wayle-config`: Custom module click fields changed from `String` to `ClickAction`
- `wayle-shell`: New `dropdowns::custom` module with `CustomDropdown` and `PickerSection` components
- `wayle-shell`: Registry fallback handles `"custom:*"` names, pre-warms custom dropdowns
- `wayle-shell`: `on_next_close` registry method for post-dropdown-close callbacks
- `wayle-styling`: SCSS for custom dropdown picker reusing popover-item mixins
- `README.md`: Documentation for the new config format

wayle-pr-scale-cropped

## Test Plan

- [x] Configure a custom module with `left-click = "dropdown:custom:<name>"`
- [x] Verify the picker dropdown opens with items from `list-command`
- [x] Verify selecting an item runs `select-command` with `$WAYLE_SELECTED`
- [x] Verify the dropdown closes after selection and bar label refreshes via `on-action`
- [x] Verify plain text fallback (one item per line, no JSON)
- [x] Verify existing custom modules with shell click actions still work
- [x] Verify built-in dropdowns are unaffected
- [x] `cargo +nightly fmt --all --check` passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo test --workspace` passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,43 @@ Scroll events are debounced (50ms) so rapid scrolling doesn't fire dozens of
 commands. Set `interval-ms = 0` if you only want updates from `on-action` (no
 polling at all).
 
+Click actions can also open dropdowns. Use `"dropdown:<name>"` for built-in
+dropdowns (audio, battery, etc.) or `"dropdown:custom:<name>"` for custom
+dropdowns defined in `[dropdowns.custom.<name>]`:
+
+```toml
+[[modules.custom]]
+id = "scale"
+command = "hypr-scale.sh get"
+interval-ms = 5000
+format = "{{ text }}"
+icon-name = "ld-monitor-symbolic"
+left-click = "dropdown:custom:scale"
+on-action = "hypr-scale.sh get"
+
+[dropdowns.custom.scale]
+title = "Display Scale"
+icon = "ld-monitor-symbolic"
+width = 340
+
+[[dropdowns.custom.scale.sections]]
+type = "picker"
+list-command = "hypr-scale.sh list --json"
+select-command = 'hypr-scale.sh set "$WAYLE_SELECTED"'
+```
+
+The `list-command` outputs one item per line — either plain text or JSON:
+
+```json
+{"value": "1.5", "label": "1.5x", "subtitle": "140 DPI", "active": true}
+```
+
+JSON fields: `value` (selection key), `label`, `subtitle`, `icon`, `active`
+(shows checkmark). Plain text lines use the text as both value and label.
+
+The `select-command` receives the chosen value via `$WAYLE_SELECTED`. After
+selection the dropdown closes and `on-action` runs to refresh the bar label.
+
 ### JSON Reserved Fields
 
 When outputting JSON, these fields have special meaning:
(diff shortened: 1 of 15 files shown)


Dieterbe:
so #71 represents each option as just a string; this PR allows the same, but also allows options with title, subtitle, a custom value returned, an icon, and an "active" mark.

I like these additional UI options quite a lot (but I have no opinion on whether the dropdown definition should be in the custom module or separately as done in this PR)

I do wonder if perhaps users may want to add more and more "stuff" to their options, e.g. maybe they'll want to add images, colors, or perhaps they'll want to customize how the dropdown options should look like in a more flexible way.  It could be interesting if an external script could return a desired widget layout, but AFAIK this is not possible. The next best option is that wayle predefines the layout and the script just returns the needed values, as you did.

waltmck:
This is an interesting idea. My thinking is that for complex but generally-useful modules like this, an upstream-first approach is best (i.e., encouraging people to contribute Rust modules to the repository). A better maintainership model where contributors share the maintenance burden of their contributed modules would be needed for this to be sustainable, but this approach has worked successfully at scale in the past (for example, see `nixpkgs`).

For bespoke use-cases, I do also see the benefits of script-ability. However, shell scripts seem a pretty poor choice for this beyond the very simple currently-supported setting. How about a proper scripting language like Lua?

PR #94, opened 2026-04-04
feat: add spec file for rpm packaging

justjustie:
Sorry for ignoring the template, it didn't really fit this kind of thing.

I have written a spec file that can be used to package an RPM for Fedora, mostly adapted from the existing PKGBUILD from the AUR wayle package.

If we wanted to go a step further, we could setup an "official" wayle repo on Fedora COPR, their kinda equivalent of Arch's AUR. There are two big differences though:
- Instead of just providing build instructions like AUR, COPR builds and packages software on their servers
- Every packager practically maintains their own repo on COPR, unlike the global AUR one

The second point could be challenge, as some of wayle's optional dependencies (swww and the theming engines) are not on Fedoras official repos. This means the user would be on their own luck sourcing them from another COPR repo of their choice or compiling them from source, or we would have to package these on wayle's "official" COPR repo.

Anyway, should there be interest in officially offering a wayle package on COPR, there is a way to set up automated builds on there using a webhook triggered from GitHub. Currently, the version number would have to be bumped in the spec file manually, I think there is a way to solve that using templating but I have not figured it out yet.

To check that everything works I have set up my own test repo, so any Fedora users here can try it out for themselves:
```
#> dnf copr enable justie/wayle
#> dnf install wayle
```

PS: the spec file doesn't _have_ to be in the main repo, this just makes it easier to version bump for releases


justjustie:
I cannot for the life of me figure out how they want me to use the rpkg templates. Like this, COPR is gonna clone the repo twice I think (once to get the spec file, then again to get the actual sources) and the version still needs to be bumped manually, but it works. Let me know if there's any interest in this at all before I spend more time trying to make it pretty

justjustie:
Okay I think I got it except for the version number now

wayle.spec
```diff
@@ -0,0 +1,81 @@
+%global debug_package %{nil}
+
+Name: wayle
+Version: 0.2.1
+Release: 1%{dist}
+Summary: A configurable desktop shell for Wayland compositors
+
+License: MIT
```
basil, line 8: This is incomplete. The resulting binary statically links hundreds of crates with varied licenses.

wayle.spec
```diff
@@ -0,0 +1,81 @@
...
+
+%prep
+{{{ git_dir_setup_macro }}}
+cargo fetch --target "$(rustc -vV | sed -n 's/host: //p')"
+
+%build
+cargo build --frozen --release
+
+%install
+install -Dm755 target/release/%{name} %{buildroot}%{_bindir}/%{name}
```
basil: Trailing whitespace:

```suggestion
install -Dm755 target/release/%{name} %{buildroot}%{_bindir}/%{name}
```

wayle.spec
```diff
@@ -0,0 +1,81 @@
+%global debug_package %{nil}
+
+Name: wayle
+Version: 0.2.1
+Release: 1%{dist}
```
basil: Per https://docs.fedoraproject.org/en-US/packaging-guidelines/DistTag/

> Use of the `%{?dist}` tag is mandatory in Fedora.

```suggestion
Release: 1%{?dist}
```

wayle.spec
```diff
@@ -0,0 +1,81 @@
+%global debug_package %{nil}
```
basil, line 1: Per https://docs.fedoraproject.org/en-US/packaging-guidelines/Debuginfo/

> If you wish to disable generation […] use `%global debug_package %{nil}` in the specfile, and be sure to add a comment next to it explaining why it was done.

There is no comment explaining why this was done.

wayle.spec
```diff
@@ -0,0 +1,81 @@
...
+BuildRequires: gtk4-layer-shell-devel
+BuildRequires: gtksourceview5-devel
+BuildRequires: systemd-devel
+BuildRequires: libxkbcommon-devel
+BuildRequires: pulseaudio-libs-devel
+BuildRequires: fftw-devel
+BuildRequires: pipewire-devel
+BuildRequires: clang-devel
+BuildRequires: cargo
+BuildRequires: pkgconf-pkg-config
```
basil, line 27:
```suggestion
BuildRequires: pkgconf-pkg-config
BuildRequires: desktop-file-utils
```

(For `desktop-file-validate` below)

wayle.spec
```diff
@@ -0,0 +1,81 @@
...
+
+Requires: gtk4
+Requires: gtk4-layer-shell
+Requires: pulseaudio-libs
+Requires: fftw
+Requires: pipewire-libs
+Requires: libglvnd-gles
```
basil, line 35:
```suggestion
Requires: libglvnd-gles
Requires: hicolor-icon-theme
```

Mandatory whenever a package ships files under `%{_datadir}/icons/hicolor/`, so the icon cache triggers fire.

wayle.spec
```diff
@@ -0,0 +1,81 @@
...
+
+%build
+cargo build --frozen --release
+
+%install
+install -Dm755 target/release/%{name} %{buildroot}%{_bindir}/%{name} 
+install -Dm755 target/release/%{name}-settings %{buildroot}%{_bindir}/%{name}-settings
+
+install -dm755 %{buildroot}%{_datadir}/%{name}/icons
+cp -r resources/icons/hicolor %{buildroot}%{_datadir}/%{name}/icons
```
basil: Use canonical directory for high color icons:

```suggestion
install -d %{buildroot}%{_datadir}/icons/hicolor/scalable/actions
install -m 0644 resources/icons/hicolor/scalable/actions/*.svg %{buildroot}%{_datadir}/icons/hicolor/scalable/actions/
```

wayle.spec
```diff
@@ -0,0 +1,81 @@
...
+target/release/%{name} completions bash > wayle.bash
+target/release/%{name} completions zsh > _wayle
+target/release/%{name} completions fish > wayle.fish
+
+install -Dm644 wayle.bash %{buildroot}%{_datadir}/bash-completion/completions/wayle
+install -Dm644 _wayle %{buildroot}%{_datadir}/zsh/site-functions/_wayle
+install -Dm644 wayle.fish %{buildroot}%{_datadir}/fish/vendor_completions.d/wayle.fish
+
+install -Dm644 resources/wayle.service %{buildroot}%{_userunitdir}/wayle.service
+install -Dm644 resources/com.wayle.settings.desktop %{buildroot}%{_datadir}/applications/com.wayle.settings.desktop
```
basil, line 69: Per https://docs.fedoraproject.org/en-US/packaging-guidelines/#_desktop_files

```suggestion
install -Dm644 resources/com.wayle.settings.desktop %{buildroot}%{_datadir}/applications/com.wayle.settings.desktop
desktop-file-validate %{buildroot}%{_datadir}/applications/com.wayle.settings.desktop
```

wayle.spec
```diff
@@ -0,0 +1,81 @@
...
+install -Dm644 resources/wayle-settings.svg %{buildroot}%{_datadir}/icons/hicolor/scalable/apps/wayle-settings.svg
+
+%files
+%{_bindir}/%{name}
+%{_bindir}/%{name}-settings
+%{_datadir}/%{name}/
```
basil: Use canonical directory for high color icons:

```suggestion
%{_datadir}/icons/hicolor/scalable/actions/*.svg
```

basil (changes requested):

wayle.spec
justjustie: Thank you for the review! I have never packaged before and saw this as opportunity to learn.

I have looked into the licensing issue and found the `%cargo_license` macro, as far as I understand we can't use that though since we can't use `%cargo_prep`/`%cargo_build` either because we have to build online. Any insight here is greatly appreciated!

basil: The main thing that macro produces is output like this for you to paste into the spec:

```
# Apache-2.0
# MIT
# MIT OR Apache-2.0
License:        Apache-2.0 AND MIT
```

Here each comment is the license of a statically-linked dependency, and the final license is the logical "and" of the license of the product and all its dependencies.

Jas-SinghFSU:
Closing due to #121

PR #97, opened 2026-04-06
fix(memory): memory growth app side

### LINKED ISSUE #95: Memory leak / RAM increasing
### Quick check

### Description

I don't know if I'm the only one experiencing this behavior, but when I was testing my development environment, I was convinced I had introduced a memory leak, so I left my build bar with the latest build from the master branch to see what would happen, and I got the same result

Of what I can see it's maybe related to notifications or media player

Image

Image

Image

### Steps to reproduce

1. Launch Wayle
2. Wait a bit and let the bar do its work

---
I can reproduce especially when I switch music on the media player of waybar and this triggered a notification at same time (related or not I don't know at moment)

### Expected behavior

Observe the RAM growing

### Wayle version

0.1.2

### Distro

Arch linux

### Compositor and version

Hyprland 0.54.3 built from branch v0.54.3 at commit 521ece463c4a9d3d128670688a34756805a4328f clean (version: bump to 0.54.3).

### Display setup

1x 5120x1440 @ scale 1

### Relevant config

```toml

```

### Logs

```shell

```

### Affected area

Bar

Times-Z:
## Objective

Resolve part of memory growth in the bar shell reported in issue #95, particularly when notifications and media player / systray activity are present
The investigation used [heaptrack](https://github.com/KDE/heaptrack) profiling across multiple runs (only in release mode) to isolate leaks and validate fixes

## Solutions

Four targeted fixes addressed the major memory leak vectors in application code:

1. **Systray CSS Provider**
   - Prevented repeatedly attaching the same CSS provider on every icon update
   - Added `icon_color_provider_attached` flag to `SystrayItem` to attach the provider only once

2. **Notification Icon Loading**
   - Replaced full-size file icon loading with scaled texture loading for both popup cards (64px) and dropdown items (48px)
   - Added `load_scaled_file_icon()` helper using `gdk_pixbuf::Pixbuf::from_file_at_scale` + `Texture::for_pixbuf`
   - Added `gdk-pixbuf` workspace dependency

3. **Systray Icon Deduplication**
   - Added `IconSignature` enum to track when icons have actually changed (vs. redundant updates)
   - `update_icon()` now compares signature and skips apply if unchanged
   - File-based systray icons are now loaded at scaled resolution instead of full-size
   - Clear image state before switching sources to help GObject release old texture references

4. **Bar Button Threshold Styling**
   - Added deduplication check so CSS is reloaded only when `ThresholdColors` actually change
   - Prevents thousands of redundant `gtk_css_provider_load_from_string` calls per session from threshold polling

**Validation results from heaptrack profiling:**
- Notification leak stacks (previously 13.11M) are no longer in top results
- Systray repeated icon update leak path is no longer a growth source
- Release builds validated after each fix

**Remaining allocations** (not actionable in this repo):
- ~4.9M: GTK-internal texture caching in snapshot chain (no Rust frame)
- ~4.3M: `wayle_systray` external crate icon pixmap D-Bus polling (zvariant/zbus allocations)
- ~4.1M: Tracing initialization (intentional `mem::forget(_guard)` for log guard lifetime)

## Test Plan

- [x] `cargo check -p wayle-shell` passes after each fix
- [x] `cargo build --release -p wayle` compiles successfully
- [x] heaptrack profiling validates leak removal across multiple sessions
- [x] Ensure `cargo fmt` and `cargo clippy --workspace` pass before merge

---

## Remaining Work / Future Improvements

| Leak Source | Size | Possible Action |
|---|---|---|
| wayle-systray icon pixmap polling | ~4.25M | Patch wayle-systray to debounce icon updates, or cache decoded pixmaps |
| GTK internal texture cache | ~4.92M | Profile GTK teardown or investigate custom texture lifecycle management |
| Tracing initialization | ~4.10M | Intentional (mem::forget for log guard); only fixable by removing logging |
| Pango/PangoFT2 font cache | ~2.4M | One-time cost during NetworkDropdown startup (acceptable) |
| PipeWire audio context | ~430K | One-time cost during CAVA module startup (acceptable) |
| sysinfo CPU temp file reads | ~131K | Consider caching temperature sensor reads in external crate |

Fixes #95


crates/wayle-shell/src/shell/notification_popup/card/methods.rs
```diff
@@ -35,8 +39,12 @@ impl NotificationPopupCard {
...
 
             ResolvedIcon::File(path) => {
-                icon.set_from_file(Some(path));
-                icon_container.add_css_class("file-icon");
+                if let Some(texture) = load_scaled_file_icon(path, POPUP_ICON_TEXTURE_SIZE_PX) {
+                    icon.set_paintable(Some(&texture));
+                    icon_container.add_css_class("file-icon");
+                } else {
+                    icon.set_icon_name(Some("ld-bell-symbolic"));
+                }
```
Jas-SinghFSU:
```suggestion
                let path = path.clone();
                let icon = icon.clone();
                let icon_container = icon_container.clone();

                spawn_local(async move {
                    let texture = gio::spawn_blocking(move || {
                        load_scaled_file_icon(&path, DROPDOWN_ICON_TEXTURE_SIZE_PX)
                    })
                    .await
                    .ok()
                    .flatten();

                    let Some(texture) = texture else {
                        icon.set_icon_name(Some("ld-bell-symbolic"));
                        return;
                    };

                    icon.set_paintable(Some(&texture));
                    icon_container.add_css_class("file-icon");
                });
```
Since we're ultimately calling `Pixbuf::from_file_at_scale` which could take some time (for large images), let's move it off the main gtk thread so it doesn't block the GTK4 main loop. Otherwise the shell would freeze while processing a large image.

crates/wayle-shell/src/shell/bar/modules/systray/item/methods.rs
```diff
@@ -182,20 +218,29 @@ impl SystrayItem {
...
             #[allow(deprecated)]
-            image.style_context().remove_provider(&provider);
+            image.style_context().remove_provider(provider);
+            self.icon_color_provider_attached = false;
         }
     }
 
     fn apply_icon(&self, image: &gtk::Image, icon_name: Option<&str>) {
+        image.set_icon_name(None);
+        image.set_paintable(None::<&gdk::Texture>);
```
Jas-SinghFSU:
```suggestion
        image.clear();
```
Could we simplify to this instead?

Jas-SinghFSU (changes requested):

Jas-SinghFSU:
Some merge conflicts and suggested changes above when you get a chance. Looks good otherwise.

Jas-SinghFSU (approved):

Jas-SinghFSU:
Sorry for being late to get to this, thank you for resolving the comments!

Times-Z:
> Sorry for being late to get to this, thank you for resolving the comments!

No worries It's ok I saw the reason on discord and this is totally understandable.
Thanks you !

PR #100, opened 2026-04-09
feat: add pomodoro module

Dieterbe:
Basic pomodoro module.
image

to operate:
- in dropdown, you can choose between work / short break / long break, and start the timer
- even with dropdown closed, you can pause/start the timer (right click by default)
- after each mode completes, auto-advances (between work and short and long breaks) as configured
- icon changes based on timer (stopped=default, started and paused)
- color of label and icon changes based on current mode via scss

so once you know how it works, it shouldn't even require opening the dropdown to operate.

Remarks:
1) the timer countdown freezes on the bar module when dropdown is open, this was actually due to a deliberate change on master i believe
2) no notifications yet, not sure i would actually want this. personally i'm happy with the more subtle color changes
3) no integration with calendar, todo list systems etc. something to explore in the future

### DIFF
--- a/crates/wayle-config/src/schemas/bar/types/mod.rs
+++ b/crates/wayle-config/src/schemas/bar/types/mod.rs
@@ -196,6 +196,8 @@ pub enum BarModule {
     Netstat,
     /// Notification center button.
     Notifications,
+    /// Pomodoro timer.
+    Pomodoro,
     /// Power menu button.
     Power,
     /// RAM usage indicator.
@@ -261,6 +263,7 @@ impl BarModule {
             Self::Network => "network",
             Self::Netstat => "netstat",
             Self::Notifications => "notifications",
+            Self::Pomodoro => "pomodoro",
             Self::Power => "power",
             Self::Ram => "ram",
             Self::Separator => "separator",
@@ -293,6 +296,7 @@ impl BarModule {
             "network" => Self::Network,
             "netstat" => Self::Netstat,
             "notifications" => Self::Notifications,
+            "pomodoro" => Self::Pomodoro,
             "power" => Self::Power,
             "ram" => Self::Ram,
             "separator" => Self::Separator,
@@ -376,6 +380,7 @@ const BUILTIN_MODULES: &[&str] = &[
     "netstat",
     "network",
     "notifications",
+    "pomodoro",
     "power",
     "ram",
     "separator",
--- a/crates/wayle-config/src/schemas/modules/mod.rs
+++ b/crates/wayle-config/src/schemas/modules/mod.rs
@@ -18,6 +18,7 @@ mod netstat;
 mod network;
 /// Notification module configuration and popup types.
 pub mod notification;
+mod pomodoro;
 mod power;
 mod ram;
 mod separator;
@@ -53,6 +54,7 @@ pub use notification::{
     IconSource, NotificationConfig, PopupCloseBehavior, PopupMonitor, PopupPosition, StackingOrder,
     UrgencyBarThreshold,
 };
+pub use pomodoro::PomodoroConfig;
 pub use power::PowerConfig;
 pub use ram::RamConfig;
 pub use separator::SeparatorConfig;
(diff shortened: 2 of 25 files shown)


Jas-SinghFSU:
This is awesome. Can we move the play button below the timer so that the timer is centered? Also I think a text based button might look nicer (though this is subjective); I'm looking at this as a reference https://pomofocus.io/.

Additionally what are your thoughts on having a reset? If you start a timer and wan to reset (misclick or other reason), is that an option that's available? What is the current interaction in that case?

Dieterbe:
> This is awesome. Can we move the play button below the timer so that the timer is centered? Also I think a text based button might look nicer (though this is subjective); I'm looking at this as a reference https://pomofocus.io/.

i don't think we should sweat the design details yet.  we will surely tweak the functionality and hash out some common flows, which will inform a better design. my personal pomodoro usage is quite simplistic. I propose we get this merged, so we get more real user feedback and then iterate based on that.

FWIW i kinda like current design (it also has minimal mouse movement between the first click on the module and the subsequent click on the start button, though not sure how important that is given you can start/pause from the module itself too) but am also open to any suggestions. I did just notice i failed to center it correctly vertically (it sits a few pixels too high)
I like the symbolic button, and if anything, i would rather make it more clear that the play/pause glyph sits on a clickable button rather than just showing the glyph as we do now.


> Additionally what are your thoughts on having a reset? If you start a timer and wan to reset (misclick or other reason), is that an option that's available? What is the current interaction in that case?

can you describe what you mean, especially with "other reason"?

with reset, do you mean:
a) go back to previous state (go back to previous mode and partially progressed time) -> this seems like the correct thing to do in case of an accidental misclick, but honestly misclicking seems such a rare case i don't think we need to accommodate it.  this is currently not supported
b) abort the timer (fill it back up) and retain the mode -> for this you currently have to click on another mode and then back to the same one. i don't think clicking the current mode should reset the timer, therefore the only "solution" here would be to add another button to do the reset, but this also seems like a contrived scenario, not worth adding a separate button for.  if you start a work timer and then decide you're not ready to work yet, you can simply pause and resume later; your timer will start from 19:58 instead of 20:00, which seems like a non-issue.. what is the actual use case or problem here?

Jas-SinghFSU:
> i don't think we should sweat the design details yet.

Fair and we don't have to. These are just small design nit picks which I think are worth base-lining from the start. Center aligned text to me seems more consistent (hence moving the button below) and emphasizes the separation between information and interaction more clearly. Additionally if we look at most online pomodoro timers, they employ the same patterns - hence more familiarity for the users.
1. https://pomofocus.io/
2. https://studywithme.io/aesthetic-pomodoro-timer/
3. https://www.morgen.so/pomodoro-timer
4. https://www.tomatotimers.com/
5. https://www.toptal.com/project-managers/tomato-timer
6. https://pomodorokitty.com/

> go back to previous state (go back to previous mode and partially progressed time) -> this seems like the correct thing to do in case of an accidental misclick, but honestly misclicking seems such a rare case i don't think we need to accommodate it. this is currently not supported

The justification is moreso that the same behavior is exposed, but via extra clicks (click another mode and back). And general UX best practices suggests minimizing clicks for users to get to the desired outcome. Additionally, to piggy back off of the the last point regarding existing patterns, it seems common to provide a 'reset' button; deviating from it (unless a strong reason is provided) adds friction to the UX for users coming from external pomodoro timers without the benefit.

> what is the actual use case or problem here?
A few I can think of off the top of my head:
1. A task change - you've finished your task and aren't ready a break
2. An interruption - a phone call, a last minute meeting invite, etc. - requiring a session reset
3. A misclick as mentioned above

Of course there's no user feedback here but IMO it would be a solid baseline feature to add (and maybe even allow disabling via config).

Dieterbe:
> emphasizes the separation between information and interaction more clearly. Additionally if we look at most online pomodoro timers, they employ the same patterns - hence more familiarity for the users.

good point. also https://pomodor.app/timer
seems very common to have a start and reset button at least [under the timer]. some also add a "skip" or "config" button. once you have more than one button, putting them under the timer certainly makes more sense.  the "skip" button seems to also address your case of "i finished but i don't want to break yet", in this case you could skip the break.

I'll go ahead and rebase/merge and put the start/stop button + a reset button under the timer. i'll hold off on the "skip" button, since you can also achieve the same by clicking the desired mode and starting the timer.

it also seems most of them put the mode on top, above the timer, i'll do that too.

that should be a good baseline for merging and getting further feedback then

Issue #42, opened 2026-03-21
Feat: Desktop Icon As Fallback before Fallback Icon in Workspaces


alexankitty:
I think it would be useful to allow using the original .desktop icon as a fallback if one is present if there is no symbolic icon available. It should be a config option though as it would allow for breaking UI consistency. Then if for some reason the .desktop icon isn't present use the fallback icon.

PerchunPak:
For reference, FreeDesktop specification:
https://specifications.freedesktop.org/icon-theme/latest/

This is especially useful for niche programs that are not popular enough to have their icon lucide/other sources.

waltmck:
Please test #325 which implements fallback to installed symbolic icons (does not fallback to color icons to preserve the existing design language and avoid a mix of symbolic and color icons in the same UI).

PerchunPak:
I have patched #325 to use color icons, and it works great; I wouldn't say it adds any inconsistencies. Much better than having a stub icon

```patch
diff --git a/crates/wayle-shell/src/shell/bar/icons.rs b/crates/wayle-shell/src/shell/bar/icons.rs
index 1ae9480..2829e33 100644
--- a/crates/wayle-shell/src/shell/bar/icons.rs
+++ b/crates/wayle-shell/src/shell/bar/icons.rs
@@ -320,10 +320,7 @@ fn resolve_symbolic_desktop_icon(identifier: &str) -> Option<String> {
     gdk::Display::default()?;
 
     let icon = desktop_entry_icon(identifier)?;
-    let base = icon.strip_suffix("-symbolic").unwrap_or(&icon);
-    let symbolic = format!("{base}-symbolic");
-
-    icon_exists(&symbolic).then_some(symbolic)
+    icon_exists(&icon).then_some(icon)
 }
 
 /// Reads the `Icon=` value from an app's desktop entry — by desktop id first,
```

PR #117, opened 2026-04-17
fix: avoid using generic glob patterns for most application icons

PerchunPak:
Replace most of the glob patterns with specific classes. This is a problem, because `*tor*` (instead of `Tor Browser`) also matched something like `qbit*tor*rent`. The same thing with VSCode's `code`.

I manually launched most of the applications from `crates/wayle-config/src/schemas/modules/window_title/icons.rs`, the unchanged globs are applications that failed to launch for me. Anyway, wayle should really move to the FreeDesktop specification.

Related: https://github.com/wayle-rs/wayle/issues/42


crates/wayle-shell/src/shell/bar/icons.rs
PerchunPak, line 1: There are tests in 

https://github.com/wayle-rs/wayle/blob/ea85d960ac7a319feef1f62b008ce289ffa729c4/crates/wayle-shell/src/shell/bar/modules/hyprland_workspaces/helpers.rs#L292-L304

And I would add somewhere tests for those two false matches from description, but I am not sure putting tests to hyprland-specific file is ideal. Where should I add it?

PerchunPak:
Want to wait for https://github.com/wayle-rs/wayle/pull/118 first

PerchunPak:
@Jas-SinghFSU I don't really understand the logic of the failing test. Shouldn't it check for `si-spotify-symbolic` instead of `spotify-symbolic`?

https://github.com/wayle-rs/wayle/blob/1dc90b4f3a8f2d49f15d3460277c139e29b1834b/crates/wayle-shell/src/shell/bar/modules/media/helpers.rs#L247-L259

Jas-SinghFSU:
> @Jas-SinghFSU I don't really understand the logic of the failing test. Shouldn't it check for `si-spotify-symbolic` instead of `spotify-symbolic`?

That tests for the symbolic icons shipped with icon themes for a lot of applications. Whereas `si-spotify-symbolic` would be one installed from simple icons.

crates/wayle-shell/src/shell/bar/icons.rs
```diff
@@ -153,40 +162,40 @@ pub(crate) const DEFAULT_APP_ICON_MAP: &[(&str, &str)] = &[
...
     // Video Editors
-    ("*blender*", "si-blender-symbolic"),
+    ("blender", "si-blender-symbolic"),
     ("*davinci*", "si-davinciresolve-symbolic"),
     ("*kdenlive*", "si-kdenlive-symbolic"),
     ("*olive*", "ld-film-symbolic"),
     ("*openshot*", "ld-film-symbolic"),
     ("*pitivi*", "ld-film-symbolic"),
-    ("*shotcut*", "ld-film-symbolic"),
+    ("*shortcut*", "ld-film-symbolic"),
```
Jas-SinghFSU: This should be `shotcut` the [app](https://www.shotcut.org/download/).

crates/wayle-config/src/schemas/modules/window_title/icons.rs
```diff
@@ -4,65 +4,74 @@
...
     // Browsers
-    ("*brave*", "si-brave-symbolic"),
-    ("*chromium*", "tb-brand-chrome-symbolic"),
-    ("*firefox*", "si-firefox-symbolic"),
-    ("*floorp*", "si-floorp-symbolic"),
-    ("*google-chrome*", "si-googlechrome-symbolic"),
-    ("*librewolf*", "si-librewolf-symbolic"),
+    ("brave-browser", "si-brave-symbolic"),
+    ("chromium-browser", "tb-brand-chrome-symbolic"),
+    ("firefox", "si-firefox-symbolic"),
```
Jas-SinghFSU: I imagine this would fail `org.mozilla.firefox`? Perhaps the right approach is somewhere in between where we introduce globs for apps that can have variants such as app, app-nightly, app-dev, app-xyz... so we can just do `app` and `app-*`.

PerchunPak: Oh, this file is for matching on Dbus names. Reverted changes then, as I don't want to launch every app again

PR #128, opened 2026-04-19
docs: document how to install on NixOS

PerchunPak:
Will wait for Wayle to hit nixos-unstable before undrafting it (when `nixos-unstable` will get a green label on https://status.nixos.org/)


PerchunPak:
Wayle is now in `nixos-unstable` and the home-manager module has also been merged just now - undrafting this

PR #152, opened 2026-04-28
feat: support workspace names as keys in workspace-map (#151)

### LINKED ISSUE #151: Support workspace names as keys
### Quick check

### Problem or use case

This relates to: https://github.com/Jas-SinghFSU/HyprPanel/issues/1190

I would like to be able to have numbered workspaces by default, while allowing configurable icons for special workspaces to enforce applications to specific workspaces.

Special workspaces (special:discord, special:spotify, etc.) get negative numeric IDs assigned by Hyprland at runtime based on creation order. The order is non-deterministic — whichever app opens first gets -98, the next gets -97. This makes workspace-map numeric keys unreliable for special workspaces: icons shuffle after most reboots.

### Proposed solution

Allow workspace-map to accept workspace name strings as keys in addition to numeric IDs:

[modules.hyprland-workspaces.workspace-map]
  "special:discord" = { icon = "󰙯", color = "#7289da" }
  "special:spotify" = { icon = "󰓇", color = "#1db954" }
  "special:steam"   = { icon = "󰓓", color = "#c6d4df" }

Name-based keys are matched against the workspace name reported by Hyprland. They are checked before numeric ID keys, so a name entry takes priority when both could match. Existing numeric key behaviour is unchanged.

I will create a PR for this change an link it to the issue, it shouldn't a large change and should follow nicely from HyprPanel change made previously :).

Feel free to do with it what you like, I am not precious as I can just create my own fork for now.

### Alternatives considered

- Forcing consistent workspace creation order at startup via a script — doesn't work reliably because Hyprland destroys empty special workspaces before apps finish loading.
- display-mode = "icon" with app icons — shows correct per-app icons but removes numbers from regular workspaces, which is an unwanted trade-off.

### Related area

Configuration

AlexTrebs:
Fixes #151

Before:
After initial load you could set the applications to a negative id. The issue arrises when you close that app and reopen the applications to the negative ids now do not match up:
image

After the changes you can map icons to the workspace name also:
image

Configuration:
```toml
[modules.hyprland-workspaces.workspace-map."special:tilde"]
icon = "utilities-terminal"
color = "#fab387"

[modules.hyprland-workspaces.workspace-map."special:spotify"]
icon = "spotify"
color = "#1db954"

[modules.hyprland-workspaces.workspace-map."special:discord"]
icon = "discord"
color = "#7289da"

[modules.hyprland-workspaces.workspace-map."special:steam"]
icon = "steam"
color = "#c6d4df"
```

## Objective

Special workspaces get negative numeric IDs assigned by Hyprland at runtime based on creation order, which is non-deterministic. This makes `workspace-map` numeric keys unreliable — icons shuffle after most reboots.

## Solution

Change `WorkspaceMap` internal key type from `i32` to `String`. Keys that parse as integers are matched against the workspace's numeric ID; all other keys are matched against the workspace name reported by Hyprland. Name-based keys are checked first, so they take priority over numeric keys.

Also adds a `workspace-name-{name}` CSS class to each workspace button so that name-based color overrides in `styling.rs` are applied correctly.

Existing numeric key behaviour is unchanged.

## Test Plan
- [ ] Numeric ID keys still resolve correctly
- [ ] Name-based keys resolve and take priority over numeric keys for the same workspace
- [ ] Color overrides apply correctly via the new `workspace-name-*` CSS class
- [ ] `cargo fmt` and `cargo clippy --workspace` pass

### DIFF
--- a/crates/wayle-config/src/schemas/modules/hyprland_workspaces/mod.rs
+++ b/crates/wayle-config/src/schemas/modules/hyprland_workspaces/mod.rs
@@ -100,37 +100,44 @@ pub struct WorkspaceStyle {
     pub color: Option<ColorValue>,
 }
 
-/// Per-workspace icon and color overrides, keyed by workspace ID.
+/// Per-workspace icon and color overrides, keyed by workspace ID or name.
 ///
-/// TOML table keys are always strings, so `"1"` parses into the workspace
-/// with ID `1`. Negative IDs refer to Hyprland's special workspaces. Keys
-/// that don't appear in the map fall back to the default behaviour set by
-/// [`HyprlandWorkspacesConfig::display_mode`].
+/// TOML table keys are always strings. Integer strings (`"1"`, `"-98"`) are
+/// matched against the workspace's numeric ID; any other string is matched
+/// against the workspace's name as reported by Hyprland.
+///
+/// Name-based keys are checked first, then numeric ID, so a name entry
+/// takes priority if both happen to match the same workspace.
+///
+/// Using name-based keys for special workspaces keeps icons and colors stable
+/// even when Hyprland reassigns the underlying numeric ID (which happens when
+/// a scratchpad is closed and reopened within the same session).
 ///
 /// ## Examples
 ///
 /// ```toml
 /// [modules.hyprland-workspaces.workspace-map]
-/// # Whole entry on one line with an inline table
+/// # Numeric ID keys (classic behaviour)
 /// 1 = { icon = "ld-globe-symbolic", color = "#4a90d9" }
 /// 2 = { icon = "ld-terminal-symbolic" }
-/// 3 = { icon = "ld-code-symbolic", color = "accent" }
 ///
-/// # Or spread the entry across its own subtable
-/// [modules.hyprland-workspaces.workspace-map.4]
-/// icon = "ld-message-square-symbolic"
-/// color = "status-success"
-///
-/// # Negative IDs target Hyprland special workspaces
-/// [modules.hyprland-workspaces.workspace-map.-99]
-/// icon = "ld-scratch-symbolic"
+/// # Name-based keys — stable across sessions and reopens
+/// "special:tilde"   = { icon = "utilities-terminal", color = "#fab387" }
+/// "special:spotify" = { icon = "spotify", color = "#1db954" }
 /// ```
 #[derive(Debug, Clone, Default, PartialEq, JsonSchema)]
 #[schemars(transparent)]
-pub struct WorkspaceMap(HashMap<i32, WorkspaceStyle>);
+pub struct WorkspaceMap(HashMap<String, WorkspaceStyle>);
+
+impl WorkspaceMap {
+    /// Look up a workspace style by name first, falling back to numeric ID.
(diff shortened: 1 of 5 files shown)


crates/wayle-config/src/schemas/modules/hyprland_workspaces/mod.rs
```diff
@@ -100,46 +100,53 @@ pub struct WorkspaceStyle {
...
+/// "special:spotify" = { icon = "spotify", color = "#1db954" }
 /// ```
 #[derive(Debug, Clone, Default, PartialEq, JsonSchema)]
 #[schemars(transparent)]
-pub struct WorkspaceMap(HashMap<i32, WorkspaceStyle>);
+pub struct WorkspaceMap(HashMap<String, WorkspaceStyle>);
+
+impl WorkspaceMap {
+    /// Look up a workspace style by name first, falling back to numeric ID.
+    pub fn lookup(&self, id: i32, name: &str) -> Option<&WorkspaceStyle> {
```
Jas-SinghFSU:
```suggestion
    pub fn lookup(&self, id: i64, name: &str) -> Option<&WorkspaceStyle> {
```
We can make this i64 so we don't have to do the i32 conversion dance at the call site.

crates/wayle-shell/src/shell/bar/modules/hyprland_workspaces/button/mod.rs
```diff
@@ -256,10 +258,10 @@ pub(crate) fn build_button_init(
     urgent_addresses: HashSet<Address>,
 ) -> WorkspaceButtonInit {
     let workspace_map = config.workspace_map.get();
-    let mapped_icon = i32::try_from(ctx.id)
+    let mapped_style = i32::try_from(ctx.id)
         .ok()
-        .and_then(|style_id| workspace_map.get(&style_id))
-        .and_then(|style| style.icon.clone());
+        .and_then(|id| workspace_map.lookup(id, ctx.name));
```
Jas-SinghFSU:
```suggestion
    let mapped_style = workspace_map.lookup(ctx.id, ctx.name);
```

Jas-SinghFSU (changes requested):
Looks good otherwise! Thanks for making the contribution ❤️

AlexTrebs:
Applied those suggestions and cleaned up PR :)

Jas-SinghFSU:
A few conflicts but looks good otherwise. The HashMaps on some of these mappings were updated to BTreeMaps to preserve ordering. Should be simple conflict resolution, lgtm otherwise. Thanks for making the changes!

AlexTrebs:
> A few conflicts but looks good otherwise. The HashMaps on some of these mappings were updated to BTreeMaps to preserve ordering. Should be simple conflict resolution, lgtm otherwise. Thanks for making the changes!

Sorry this took a while! All resolved

Jas-SinghFSU:
No worries, thank you! Will take a look.

Jas-SinghFSU:
https://github.com/wayle-rs/wayle/actions/runs/30170994923/job/89712323662?pr=152

Some linter failures as well.

AlexTrebs:
> https://github.com/wayle-rs/wayle/actions/runs/30170994923/job/89712323662?pr=152
> 
> Some linter failures as well.

All done :)

PR #165, opened 2026-05-01
feat: Media dropdown add hours to duration format if media is longer than 1 hour

5ila5:
Media Dropdown shows there digit minute count instead of usual `hh:mm:ss` format if the media is longer than one hour.

This is a simple fix:
- If the duration is longer then an hour, we format with `hh:mm:ss`
- If the duration is shorter than an hour, we format with `mm:ss`.

I added this simple fix and an appropriate test case

One could add a setting for this, but I'd think the majority of people would prefer it this way

For Some reason I cannot attach an image right now (no clue why) but it looks like expected. For my test YouTube video shows 
3:26:55 / 5:21:07 instead of 212:26 / 321:07


Jas-SinghFSU:
Great catch! Thanks for making the fix :heart:

PR #169, opened 2026-05-02
fix(weather): use Fluent keys for day labels in daily forecast

sim590:
## Summary

`chrono::format("%a")` always returns day abbreviations in English regardless of the user's locale. This replaces it with the existing Fluent `cal-weekday-*` keys so that day labels in the daily weather forecast are properly localized.

## Changes

- `day_label()` in `daily_forecast/helpers.rs` now uses `t!("cal-weekday-*")` instead of `date.format("%a")`
- Updated the corresponding unit test to check that the label is non-empty rather than hardcoding `"Mon"`

## Impact

This affects all non-English locales — any translation that provides `cal-weekday-*` keys will now display correctly in the weather forecast.


Jas-SinghFSU:
Hmm, if we're going to do this let's make a separate entry in the weather dropdown fluent file. The calendar uses 2 letters due to condensed size constraints which doesn't look too good on the weather module.

image

sim590:
@Jas-SinghFSU: I pushed a version with 3 letters abbreviations. Or would you prefer full words? I can adjust if you prefer that!

Jas-SinghFSU:
No, this is perfect! :tada:

PR #171, opened 2026-05-03
feat(osd): add toggle_keys config option to suppress keyboard toggle OSD

### LINKED ISSUE #91: Ignore some OSD types
### Quick check

### Problem or use case

I have binded capslock to switching keyboard layouts, so now when I press it, I always have this pop-up

Image

Related config:

```ini
# hyprland.conf
input {
  kb_layout=us,ua
  kb_options=grp:caps_toggle
}
```

### Proposed solution

I would like to disable OSD only for capslock, volume/brightness are useful

### Alternatives considered

_No response_

### Related area

OSD

v-engelowski:
Fixes #91

## Objective
Allow users to disable the OSD for keyboard toggle keys (Caps Lock, Num Lock, Scroll Lock), useful when those keys are rebound (e.g. Caps Lock -> Escape in Hyprland).

## Solution
Added a 'toggle-keys' boolean field to 'OsdConfig' (default 'true') and placed it into the toggle watcher loop so the setting is checked live on each keypress. Exposed the option in the OSD settings page with a label and description.

## Test Plan
- Rebound Caps Lock to Escape in Hyprland; confirmed OSD no longer appears with 'toggle-keys = false'
- Confirmed OSD still appears for Num Lock and Scroll Lock when 'toggle-keys = true'
- Confirmed OSD still appears for Media Keys (Mute, Volume Up/Down) when 'toggle-keys = false'
- Confirmed the setting takes effect immediately without restarting wayle
- 'cargo fmt' and 'cargo clippy --workspace' pass

Note: this is my first (real) open source contribution, so feedback on anything I've missed or could have done better is very welcome!


simonm:
Absolutely loving Wayle. It's excellent and I am one tiny step away from completely removing waybar (which I have spent countless hours tweaking).

I came to the repo to find a solution to the caps-lock triggering an OSD every time I use it. I remap caps-lock to an additional ctrl key.

Would love this PR to be merged. Anything I can do to help?

(thank you [v-engelowski](https://github.com/v-engelowski))

PerchunPak (approved):
I guess this works, but it doesn't resolve my issue. I would like to get notifications like numpad, but not about capslock.

Solution proposed in https://github.com/wayle-rs/wayle/issues/91#issuecomment-4466883360 would work perfectly even without any config change from users, but let's keep it for a separate PR.

BTW, nice code, congrats for your first contribution

Jas-SinghFSU:
Hey there! Thanks for making this PR :heart: . I think a more extensible approach would be to allow granularity in the selections so we can do:
```toml
caps_lock_show = false
num_lock_show = false
```
So on and so forth. Additionally, as mentioned above, the ability to determine previous state and not show the OSD if it hasn't changed would be useful as well to remove some of the OSD noisiness.

Issue #16, opened 2026-03-07
Some small features missing from HyprPanel


justjustie:
Just to keep track of some small features I used in HyprPanel that I noticed are not implemented in Wayle yet:
- cava module
    - configurable onclick actions (including spawning dropdowns)
    - auto hiding when no audio is playing
- calendar dropdown
    - configurable option to have the clock show seconds

***
Awesome work so far btw, can’t wait to fully migrate when dashboard and notifications are ready

Issue #15, opened 2026-03-07
Wayle network primary connection value is unknown when it should be wifi.


JohnOberhauser:
After connecting to a new network, `NetworkService.primary.get()` always gives `ConnectionType.Unknown`.  I noticed [in your shell](https://github.com/Jas-SinghFSU/wayle/blob/master/crates/wayle-shell/src/shell/bar/modules/network/mod.rs#L195) you work around this by trying wifi and wired if the connection is unknown.  Is this a network manager issue or a bug in the network crate?

PR #184, opened 2026-05-10
fix: fetch charge support info from battery device instead of DisplayDevice

sethechosenone:
## Objective
As it stands, because the battery module fetches DBus properties from DisplayDevice, charge limits always be marked as "not supported" since the DisplayDevice does not contain charge limit support information. The relevant issue is posted in wayle-services [(#15)](https://github.com/wayle-rs/wayle-services/issues/15). This is a companion PR to wayle-services [#16](https://github.com/wayle-rs/wayle-services/pull/16)

## Solution
In wayle-services, a new UPower proxy is added with an enumerate_devices() function. In the Wayle shell, the first battery device is found and used instead of DisplayDevice. This diverges from what is recommended by UPower, but will be fine for the majority of devices. It is possible to keep DisplayDevice for power information and just use individual battery devices for charge limits, however it'll require an extra design decision (I'd like some input from maintainers on this).

## Test Plan
1. Pull the changes in both wayle and wayle-services
2. In the wayle repo, add the following to Cargo.toml:
[patch.crates-io]                                            
wayle-battery = { path = "/path/to/wayle-services/wayle-battery" }
wayle-core = { path = "/path/to/wayle-services/wayle-core" }       
wayle-traits = { path = "/path/to/wayle-services/wayle-traits" }
3. Compile and run with `cargo run --bin wayle -- shell`
4. Open the battery module and verify the charge limit switch is shown

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4523,8 +4523,6 @@ dependencies = [
 [[package]]
 name = "wayle-battery"
 version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a84c4e68a9846ac0ef9bc9582595480250bfd189be6b4857965a7843766b7bb3"
 dependencies = [
  "async-trait",
  "derive_more",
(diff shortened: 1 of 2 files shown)


skifli:
Any updates on this :eyes:?

sethechosenone:
The wayle-services component got merged, but there were linter issues that I assume are going to be present here

PR #191, opened 2026-05-15
fix: workspace switching broken with Hyprland 0.55 Lua config

lapsha38:
Clicking workspace buttons stopped working after upgrading to Hyprland 0.55.

## What changed in Hyprland 0.55

Hyprland 0.55 introduced Lua-based configuration. With a Lua config active, IPC dispatch commands are no longer executed directly — they are evaluated as Lua expressions. The old format:

```
dispatch workspace 1
```

gets wrapped by Hyprland as:

```lua
return hl.dispatch(workspace 1)
```

which is a Lua syntax error, so nothing happens. This affects any tool that dispatches workspace commands via the socket, including `hyprctl` itself.

## Fix

The correct format for Hyprland 0.55+ Lua configs is:

```
dispatch hl.dsp.focus({workspace = 1})
```

The code now tries this format first. If Hyprland returns an error response (which happens on older versions or when a legacy `.conf` config is active), it falls back to the original `workspace N` format. Backward compatibility is preserved.


Times-Z:
Up ! Just migrate to lua config, and we need that !

Jas-SinghFSU:
Thank you for making this! Testing now

lapsha38:
> Thank you for making this! Testing now

nice app, u are welcome

PR #193, opened 2026-05-16
feat: add `exclusive` config option

Ape:
Introduce `bar.exclusive` to control whether the bar reserves compositor screen space. The option defaults to true to preserve existing behavior, while setting it to false makes the bar draw over windows.

This is analogous to the `exclusive` config in Waybar. Very useful when having autohide on.


Jas-SinghFSU:
Great work! Thanks for your contribution :tada:

Issue #198, opened 2026-05-18
Feature request: hide media module when no player is active


nnazdu:
### Quick check

- [x] I searched existing issues first.

### Problem or use case

Right now the media module is always visible, even when nothing is playing. This wastes bar space and can look cluttered on minimal setups.

### Proposed solution

Add a configuration option (e.g. `hide-when-empty`) to the `media` module that automatically hides it from the bar when no MPRIS player is present or all players are stopped.

### Alternatives considered

It is currently possible to approximate this behaviour with a custom module running playerctl, but that sacrifices the built-in media UI.

### Related area

Configuration

PR #201, opened 2026-05-19
feat: add visibility toggle for media module

nnazdu:
Proposed solution for #198 

## Objective

Add toggle to hide media module when no MPRIS player is detected.

## Solution

implemented an `update_visibility` method for the media module to check against config if the module should be able to hide or not. It then checks the player's playback state to assert wether it is playing or not. 

## Test Plan

Tested the defaults with the toggle disabled.
With the toggle activated, the logic works as intended.

_Note_: this is my first real contribution and my rust is academic and well, to say the least, rusty.  the logic of the changes is my work but the code was cleaned up with the help of AI.


Jas-SinghFSU:
I do like where we're going with this. Though I would suggest that we make it more configurable than a boolean.

A more extensible approach would be to implement a `visibility` setting that can take one of:
1. always = always show (default)
2. playing = show only when playing something (your implementation)
3. active = show only when a media player actually exists

This way if we need to add more variants we can without having to introduce new booleans.

nnazdu:
Thank you for the input!
I'm closing this PR and linking it to a new one since i branched out from master.

PR #223, opened 2026-05-25
feat: add visibility toggle for media module

nnazdu:
Proposed solution for #198 
Re of #201 

## Objective

Add toggle to hide media module when no MPRIS player is detected.

## Solution

implemented an `update_visibility` method for the media module to check against config if the module should be able to hide or not. It then checks the player's playback state to assert wether it is playing or not. 
Following @Jas-SinghFSU 's advice i switched from a bool to the `visibility` enum, and then adapted the logic with this change in mind.

## Test Plan

Tested the defaults with the toggle disabled.
With the toggle activated, the logic works as intended.

_Note_: this is my first real contribution and my rust is academic and well, to say the least, rusty.  the logic of the changes is my work but the code was cleaned up with the help of AI.

### DIFF
--- a/crates/wayle-config/src/schemas/modules/media/mod.rs
+++ b/crates/wayle-config/src/schemas/modules/media/mod.rs
@@ -85,6 +85,11 @@ pub struct MediaConfig {
     #[default(String::from("{{ title }} - {{ artist }}"))]
     pub format: ConfigProperty<String>,
 
+    /// Visibility mode for the module.
+    #[serde(rename = "visibility")]
+    #[default(MediaVisibility::Always)]
+    pub visibility: ConfigProperty<MediaVisibility>,
+
     /// Symbolic icon name for default mode.
     #[serde(rename = "icon-name")]
     #[default(String::from("ld-music-symbolic"))]
@@ -188,6 +193,29 @@ impl MediaIconType {
             Self::Application => "application",
             Self::SpinningDisc => "spinning-disc",
             Self::ApplicationMapped => "application-mapped",
+        }
+    }
+}
+
+/// Visibility mode for the media module.
+#[wayle_enum(default)]
+pub enum MediaVisibility {
+    /// Always show the module.
+    #[default]
+    Always,
+    /// Only show when a player is in Playing state.
+    Playing,
+    /// Only show when an active media player exists.
+    Active,
+}
+
+impl MediaVisibility {
+    /// Returns the kebab-case string representation.
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Always => "always",
+            Self::Playing => "playing",
+            Self::Active => "active",
         }
     }
 }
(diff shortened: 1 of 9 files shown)


crates/wayle-shell/src/shell/bar/modules/media/methods.rs
```diff
@@ -1,9 +1,26 @@
...
 use super::MediaModule;
 
 impl MediaModule {
+    /// Determines if the module should be visible based on config and playback state.
+    pub(super) fn update_visibility(config: &MediaConfig, media: &MediaService) -> bool {
+        match config.visibility.get() {
+            MediaVisibility::Always => true,
+            MediaVisibility::Playing => {
+                // Only show if there is an active player that is currently playing.
+                if let Some(player) = media.active_player() {
```
Jas-SinghFSU, line 14: There's a bit of an edge case here. Let's say you have Spotify and Firefox open. Firefox is playing audio, however spotify is playing nothing - and spotify is your active media player. This would hide the media module and you'd have no mechanism of controlling your currently playing media (in your non-active media player).

I think the module should stay visible as long as _any_ media player is playing audio.

crates/wayle-shell/src/shell/bar/modules/media/mod.rs
```diff
@@ -154,13 +160,22 @@ impl Component for MediaModule {
...
+
                 if let Some(player) = self.media.active_player() {
                     let label = helpers::build_label(media_config, &player);
                     self.bar_button.emit(BarButtonInput::SetLabel(label));
                     let state = player.playback_state.get();
                     Self::update_spinning_state(root, state);
                 }
             }
+            MediaCmd::VisibilityChanged => {
+                // Update visibility when the 'show-only-playing' config option is toggled.
```
Jas-SinghFSU, line 175: Nit: Remove comment, no longer relevant as that setting doesn't exist anymore.

crates/wayle-shell/src/shell/bar/modules/media/methods.rs
```diff
@@ -1,9 +1,26 @@
 use relm4::{gtk, gtk::prelude::*};
-use wayle_media::types::PlaybackState;
+use wayle_config::schemas::modules::{MediaConfig, MediaVisibility};
+use wayle_media::{MediaService, types::PlaybackState};
 
 use super::MediaModule;
 
 impl MediaModule {
+    /// Determines if the module should be visible based on config and playback state.
+    pub(super) fn update_visibility(config: &MediaConfig, media: &MediaService) -> bool {
```
Jas-SinghFSU, line 9: `update` implies command, however here we're computing a boolean. Let's use a common boolean conventions of an `is_` prefix: `is_visible`

crates/wayle-config/src/schemas/modules/media/mod.rs
```diff
@@ -192,6 +197,29 @@ impl MediaIconType {
...
+    /// Always show the module.
+    #[default]
+    Always,
+    /// Only show when a player is in Playing state.
+    Playing,
+    /// Only show when an active media player exists.
+    Active,
+}
+
+impl MediaVisibility {
```
Jas-SinghFSU, line 212: Shouldn't need this since the `wayle_enum` attribute already makes enums kebab-case.

Jas-SinghFSU (changes requested):

PR #236, opened 2026-05-28
feat: Add Ubuntu package

basil:
Tested by installing the package and starting both `wayle` and `wayle-settings` on Ubuntu 26.04.


.github/workflows/ci.yml
```diff
@@ -146,18 +146,74 @@ jobs:
             - uses: Swatinem/rust-cache@v2
 
             - name: Build workspace
-              run: cargo build --workspace
+              run: cargo build --workspace --release
```
basil, line 149: Needed because a debug build doesn't produce a working package: language files are missing and Wayle crashes on launch.

Jas-SinghFSU:
Is there a reason to not include `Suggests:` blocks in `control.in` for all the runtime deps like `upower`, `bluez`, `pipewire`, etc.? And probably similarly for the `wayle.spec.in` from the last RPM efforts?

basil:
@Jas-SinghFSU Agreed, done in commit 487f7dd.

Issue #176, opened 2026-05-07
No battery detected


liamthexpl0rer:
### Quick check

- [x] I searched existing issues first.

### Description

Hi there,
I'm currently using an Lenovo Thinkpad E16 and my battery gets not detected by wayle.
The battery icon shows a warning and the dropdown tells me, that no battery was detected.

### Steps to reproduce

1. Run `wayle panel start`
2. See the battery warning icon and "No battery detected" in dropdown

### Expected behavior

The icon should show the current an remaining capacity.

### Wayle version

0.2.1

### Distro

NixOS Unstable

### Compositor and version

Hyprland 0.54.3

### Relevant config

```toml
[modules]
[modules.battery]
icon-color = '#ffffff'
label-show = false
```

### Logs

```shell
2026-05-07T12:48:00.600284Z DEBUG zbus::object_server: Returning error: org.freedesktop.DBus.Error.UnknownObject: Unknown object '/StatusNotifierItem'
    at /build/wayle-0.2.1-vendor/source-registry-0/zbus-5.13.2/src/object_server/mod.rs:444 on tokio-runtime-worker ThreadId(4)
    in zbus::object_server::dispatch_call with msg: Msg { type: MethodCall, serial: 353, sender: UniqueName(":1.98"), path: ObjectPath("/StatusNotifierItem"), iface: InterfaceName("org.freedesktop.DBus.Properties"), member: MemberName("GetAll"), body: Str, fds: [] }, hdr: Header { primary: PrimaryHeader { endian_sig: Little, msg_type: MethodCall, flags: BitFlags<Flags>(0b0), protocol_version: 1, body_len: 31, serial_num: 353 }, fields: Fields { path: Some(ObjectPath("/StatusNotifierItem")), interface: Some(InterfaceName("org.freedesktop.DBus.Properties")), member: Some(MemberName("GetAll")), error_name: None, reply_serial: None, destination: Some(BusName::Unique(":1.99")), sender: Some(UniqueName(":1.98")), signature: Str, unix_fds: None } }
    in zbus::connection::obj_server_task
    in zbus::connection::start_object_server with started_event: None
```

### Affected area

Bar

TLJ-Piffytu:
Do you have upower installed? The battery module depends on it

liamthexpl0rer:
@TLJ-Piffytu Good to know! I've installed it and the current battery status is displayed.
Do you also know, which packages do I need for the other functions to work? The charging limit & power profile is not working.

[...]

PR #254, opened 2026-05-31
docs: document upower dependency for battery module

NemurenaiDev:
## Objective
Clarify that the battery module depends on `upower` and requires the daemon to be running with an exposed battery device.

Related to #176

## Solution
Add a warning block in `docs/config/modules/battery.md`

## Test Plan
Render the site and check the `/config/modules/battery`


Jas-SinghFSU:
Since docs are auto-generated from code, this will get overridden. Perhaps we can add it to the module docs

NemurenaiDev:
Sorry, Havent noticed... I have generated the docs using `config docs --out ./docs` after the changes.

PR #270, opened 2026-06-04
feat: add hide_when_inactive for media module

YaQia:
## Objective

I found that media is always showing even when there is no media player running, which is a little strange to me.
## Solution

I added this config for media module:
```toml
[modules.media]
hide-when-inactive = true
```
And add this config to `wayle-settings` as well:
图片

## Test Plan

Tested in cargo build (debug mode), works fine.


YaQia:
Sorry, didn't search for all PRs. #223 is already trying to fix this problem. I will close this PR.

PR #300, opened 2026-06-26
feat: add module IWD

### LINKED ISSUE #284: iwd as backend for network module.
### Quick check

### Problem or use case

I don't want network manager on my system, and don't need any of its features

### Proposed solution

have an option to just use iwd.

### Alternatives considered

_No response_

### Related area

Dropdowns (volume, network, bluetooth, etc.)

waltmck:
Added module for controlling [IWD](https://wiki.archlinux.org/title/Iwd) over its dbus interface. This was made with Claude code, but is well-tested with several iterations of bug hunting. I also tried to match the structure and style of the existing Network module as closely as possible.

Closes #284.


waltmck:
I am continuing to run this on my laptop and making some tweaks, but it is ready for review and testing by anyone who is interested.

A 5k LoC PR is understandably daunting from a maintenance burden perspective, especially if the current maintainers don't use IWD. I'm sympathetic to this, so it might be helpful to come up with a model of code ownership that would spread the work and make large contributions like this practical.

I would like to volunteer as a maintainer for this module: i.e., I promise to continually fix bugs and adapt to internal changes when necessary. If I am at some point unable to continue maintaining it, it should be very easy to remove since it is mostly self-contained: the only exception is [this change to the Dashboard UI](https://github.com/wayle-rs/wayle/pull/300/changes/c337641af9d0b5d3db3ad77193442853c986c189) (making the quick connect widget fallback to using IWD when the Network module is unavailable), but I would be happy to remove that commit from the PR if it eases any maintainability concerns.

PR #318, opened 2026-07-04
feat: --config flag to pass config.toml path to wayle shell

### LINKED ISSUE #271: Feature request: allow passing config file path with command line
### Quick check

### Problem or use case

I would like to declaratively configure Wayle on NixOS without home-manager.

### Proposed solution

Add a CLI flag `-c <PATH>` or `--config <PATH>` which tells Wayle to liik for a config file at a given path.

### Alternatives considered

Alternatively, you may look for a global configuration at `/etc/wayle/config.toml` or similar and use that if it exists.

### Related area

None

waltmck:
Closes #271. Flag only applies to `wayle shell` (not `wayle panel` etc.).

### DIFF
--- a/crates/wayle-config/src/infrastructure/service.rs
+++ b/crates/wayle-config/src/infrastructure/service.rs
@@ -1,5 +1,6 @@
 use std::{
     io,
+    path::{Path, PathBuf},
     sync::{Arc, RwLock},
 };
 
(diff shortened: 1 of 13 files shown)


waltmck:
This needs some additional work to support `wayle-settings`.

PR #328, opened 2026-07-12
feat: improved dropdown menus

### LINKED ISSUE #62: Mouse interactions wont repeat unless you move the cursor.
### Quick check

### Description

Clicking on a module then clicking again does nothing. Need to move the mouse first.

I have this settings
```
scroll-up = "wayle audio output-volume +5"
scroll-down = "wayle audio output-volume -5"
```
under the volume module and the mouse wheel will register the first time the cursor enters, but after that it sometimes changes and sometimes doesn't if I keep scrolling. if I leave the module and come back with the cursor it will change at least once.

### Steps to reproduce

mouse over a module.
click
click again

or try adding mouse wheel to the volume module and try repeat scrolls

### Expected behavior

repeat mouse actions should register without having to move the mouse

### Wayle version

0.1.0

### Distro

Linux 6.19.8-1-cachyos

### Compositor and version

Hyprland 0.54.2

### Display setup

1x 2560x1440

### Relevant config

```toml
[modules.volume]
level-icons = [
    "ld-volume-symbolic",
    "ld-volume-1-symbolic",
    "ld-volume-2-symbolic",
]
icon-muted = "ld-volume-x-symbolic"
border-show = false
border-color = "red"
icon-show = true
icon-color = "auto"
icon-bg-color = "red"
label-show = true
label-color = "red"
label-max-length = 0
button-bg-color = "bg-surface-elevated"
left-click = "dropdown:audio"
right-click = ""
middle-click = "wayle audio output-mute"
scroll-up = "wayle audio output-volume +5"
scroll-down = "wayle audio output-volume -5"
dropdown-app-icons = "mapped"

```

### Logs

```shell

```

### Affected area

Bar

### LINKED ISSUE #173: Feat: Launch dropdowns from command line
### Quick check

### Problem or use case

Would be very handy to be able to launch dropdowns from command line, so these can be bound to keyboard shortcuts (specifically showing the dropdown for notifications). 

### Proposed solution

something like
`wayle dropdown show/toggle <dropdown>`

### Alternatives considered

_No response_

### Related area

None

### LINKED ISSUE #285: Bug: no scrolling for systray dropdown menus
### Quick check

### Description

Systray icons with a long list of dropdown options overflow rather than being scrollable.

### Steps to reproduce

Open a dropdown menu with many items (I am using [mulltray](https://github.com/Andriamanitra/mulltray/)). See that scrolling doesn't work, and that the menu overflows the top of the screen as shown below:

Image

### Expected behavior

A long list of dropdown options should be functional by scrolling, without overflowing the screen.

### Wayle version

0.6.0

### Distro

NixOS 26.11 (Zokor) aarch64

### Compositor and version

Hyprland 0.55.4

### Display setup

1x 2560x1664 @ scale 1.60

### Relevant config

```toml

```

### Logs

```shell

```

### Affected area

Dropdowns (volume, network, bluetooth, etc.)

waltmck:
#285 was caused by GtkPopoverMenu not being able to scroll: so, a menu/submenu that was too tall would simply overflow the screen and become unusable. In the process of rewriting this, I ran into some issues with autohide popovers.

GTK's autohide popover is, on Wayland, a xdg_popup grab (captures pointer + keyboard). On Hyprland/wlroots that grab is inherently buggy in ways that seem to be unavoidable:
- Input freezes until the next pointer motion after an outside-click dismissal (#62).
- The dismiss-click is swallowed by the protocol, so clicking another bar button dismisses but doesn't open it.
- Nested autohide popovers (our cascading systray menu) deadlock the grab chain.

My solution is to introduce a scrim, a transparent full-screen window that intercepts mouse and keyboard input whenever a popover is open. This is similar to and inspired by the approach used by eww/AGS/Astal, which draw their popup menus as a child widget inside a transparent fullscreen window (for the same reasons, to avoid buggy autohide behavior). However, I wanted to allow users to quickly switch between dropdowns with a single click, similar to the behavior of the bar on MacOS. The eww/AGS/Astal approach would layer the scrim above the bar, making other bar buttons non-functional when a popup was open (clicking them would just close the currently open popup). 

My approach keeps dropdowns as real GTK popovers parented to bar buttons (their own surfaces), and the scrim is only a dismiss catcher, never a content host. That gives GTK-native popover positioning/flip/scroll. and one-click dropdown switching is achieved by keeping the bar layered above the scrim.

Additionally, whenever a popup is open the bar (and popup) will now be layered above even full-screen applications (again matching the behavior of MacOS). This allows CLI dispatchers (keep reading) to function even when a full-screen app is open.

Finally, in implementing the scrim-based approach I ran into several focus-grabbing race conditions. The cleanest solution to these turned out to be centralizing dropdown dispatch into a single dropdown coordinator. This made it straightforward to implement a CLI to open/close/toggle dropdowns, which I also implemented. This can be used by:
```
wayle dropdown list
```
which lists available dropdown menus; these can be opened and closed with `wayle dropdown open <name>` or `wayle dropdown toggle <name>`. By default dropdowns are opened/closed on the currently active monitor (which is determined using compositor-specific features), but as a fallback they will open/close on all monitors. The monitor can be overridden by passing `--monitor=<name>` or `--monitor=all`. Similar CLI bindings can control systray icons (`wayle systray open` and `wayle systray toggle`), with the same `--monitor` semantics.

Closes #62, closes #285, closes #173.

Note the merge conflicts with my existing PRs; see merge commit [here](https://github.com/waltmck/wayle/commit/485eb2736bad1aca358e7b86bfc86b1ffdfc7a89).

Extensively tested on Hyprland, but needs additional testing on other compositors (Sway, Niri, Mango, etc.).


justjustie:
Sounds like this would also close #173

waltmck:
> Sounds like this would also close #173 

Yep it does, I missed that one. Added it to the description.

crates/wayle-config/src/infrastructure/service.rs
```diff
@@ -20,6 +20,30 @@ use crate::{
     Config, ExtractRuntimeValues, infrastructure::themes::utils::load_themes,
 };
 
+/// Config keys removed in a past release, as `(dotted-path, note)`. A stale value
+/// left in a user's config does nothing, but silently ignoring it is confusing —
+/// warn at load so the user knows to delete it. (Renamed — not removed — keys use
+/// the derive's `#[wayle(deprecated_alias)]` instead, which still applies them.)
+const REMOVED_CONFIG_KEYS: &[(&str, &str)] = &[(
```
Jas-SinghFSU: Can we instead add a new derived attribute above the config option in the struct instead? Example:
```rust
#[wayle(deprecated("dropdowns now always dismiss on outside click"))]                                                                                          
pub dropdown_autohide: Removed<bool>,  
```

This removes the potential for mistakes (invalid paths, typos, etc) and makes it immediately visible when reading the struct.

Jas-SinghFSU:
Still in the process of reviewing, this is quite a chonker so I need to understand the intent, architecture and code - may take a bit.

Though one regression I noticed is that, on multi-monitor setups previously, the dropdown would hide if we clicked somewhere on another monitor. Here's an example scenario in which you have 2 monitors DP-1, DP-2:
1. You click the audio dropdown in DP-1 and it opens the dropdown
2. You move your mouse to DP-2 and click somewhere
3. The dropdown in DP-1 now closes due to that click

Now, that dropdown remains open unless you click somewhere on the monitor in which the dropdown is open. Which is a behavioral regression. Is there some way to tackle this, perhaps with cross monitor shims? We do have to be aware of not adding more complexity to satisfy this though. We'll have to make that tradeoff and whether this regression is worth the one click dropdown switching.

waltmck:
> Still in the process of reviewing, this is quite a chonker so I need to understand the intent, architecture and code - may take a bit.

Of course, take your time! Thanks for the review.

> Though one regression I noticed is that, on multi-monitor setups previously, the dropdown would hide if we clicked somewhere on another monitor. Here's an example scenario in which you have 2 monitors DP-1, DP-2:
> 
>     1. You click the audio dropdown in DP-1 and it opens the dropdown
> 
>     2. You move your mouse to DP-2 and click somewhere
> 
>     3. The dropdown in DP-1 now closes due to that click
> 
> 
> Now, that dropdown remains open unless you click somewhere on the monitor in which the dropdown is open. Which is a behavioral regression. Is there some way to tackle this, perhaps with cross monitor shims? We do have to be aware of not adding more complexity to satisfy this though. We'll have to make that tradeoff and whether this regression is worth the one click dropdown switching.

My choice to change this behavior was actually intentional. The things that made me lean toward the per-monitor dropdown design were
1. The new behavior matches MacOS's dropdowns, which will likely be familiar to many people.
2. Global click-to-dismiss would lead to strange behavior with remote desktops: opening a menu on the remote desktop's virtual monitor would also open the scrim on the physical monitor, which seems counterintuitive to me.
3. The scrim prevents keyboard input to a visible window. If a dropdown were opened on monitor 1 and this opened a scrim on monitor 2, keyboard inputs intended for a visible window on monitor 2 would be lost. This is counterintuitive because there is not visual indication on monitor 2 that the keyboard inputs are being swallowed (since there is no dropdown open). This behavior was also a UI shortcoming of the old approach.

If you would like a global click-to-dismiss I do think it can be implemented without too much additional complexity. A few follow-up questions:
1. Should opening a dropdown on monitor 1 close other dropdowns that are open on monitor 2?
2. Should opening a dropdown on monitor 1 also open the same dropdown on monitor 2, or just the scrim?
3. If no to (2): how should we visually indicate on monitor 2 that keyboard inputs are being swallowed?
4. If no to (2): should keyboard inputs on monitor 2's scrim control monitor 1's dropdowns?

crates/wayle-config/src/infrastructure/service.rs
waltmck: Done, thanks!

PR #331, opened 2026-07-15
feat: nerd font text-icon support for bar buttons and workspaces

ExoAcox:
## Objective

Bar button and workspace icons are currently limited to GTK symbolic icon names. Nerd Font glyphs (e.g., , , ) cannot be used because they are treated as icon names and fail silently. This adds first-class Nerd Font text-icon support.

## Solution

Detect Nerd Font glyphs and render them as text labels instead of failing as unrecognized GTK icon names. No config changes needed, detection is automatic. Any icon value that is non-ASCII or very short (≤2 chars) renders as text.

BarButton component now conditionally renders either a Gtk::Label (text-icon) or Gtk::Image (symbolic), with adjustable scale config across all button variants. Workspace modules (hyprland, niri, mango) receive the same treatment.

```
[modules.window-title.icon-mappings]
kitty = ""

[modules.hyprland-workspaces.app-icon-map]
firefox = "󰈹"

[bar]
button-icon-size = 1
button-text-icon-size = 1.5

[modules.hyprland-workspaces]
icon-size = 1
text-icon-size = 1.5
```

## Screenshot

Before
2026-07-15-161528_hyprshot

After
2026-07-15-161406_hyprshot

## Test Plan

Cargo test - pass
Cargo build - clean
Verified live on Hyprland


Jas-SinghFSU:
Hey there, thank you for making this PR! Having had nerd-fonts enabled before in Hyprpanel, I'm not too inclined to introduce them back into Wayle. There were many issues I found w/ them; the most prominent being that they were extremely difficult to align properly.

Since they're fonts, depending on your font family, they would often not be center aligned in the container unless you set up the proper size or custom margins for them. Even with a consistent font, there would regularly be consistent asymmetry in their aligment (fix one, but now others are mis-aligned):
1. https://github.com/Jas-SinghFSU/HyprPanel/issues/96
2. https://github.com/Jas-SinghFSU/HyprPanel/issues/546
3. https://github.com/Jas-SinghFSU/HyprPanel/issues/519

Personally I'm not the biggest fan of nerd-fonts where SVG icons are first class. I feel that introducing this back will put is back to square one where we'll be seeing issues opened regarding misaligned icons and icon spacing issues. For that reason, I'll be closing this MR, sorry :(

